PR #385, opened 2014-12-22
TEIID-2816: HBase Connector/Translator

kylinsoong:
The commit mainly implement Query(GROUP BY,  ORDER BY, DISTINCT, LIMIT and Condition like =, <, >, BETWEEN, LIKE, IN, etc ),  Insert(BatchedInsert).


shawkins:
What is the rationale for using a custom resource adapter rather than relying on the container?

The translator also seems to duplicate the JDBC translator code, could we just have it extend the JDBC logic instead?

kylinsoong:
I agree with you, use container datasource can simplify the logic, I am doing this and also reusing the JDBC logic in translator.

shawkins:
Yes, that looks good Kylin.  The remaining changes would be:

add a release note in build/kits/docs/teiid/teiid-releasenotes.html

update integration test - from test-integration/common run $mvn -P arqullian-tests clean install
there should be a failure with the number of translators avalable.

based upon a recent issue it would be good to replace tabs with spaces for consistency.  if you are using eclipse there is a setting to always expand tabs with spaces.

use the supresswarnings annotation for non-localized strings in tests or the //$NON-NLS-X$ comment in source

and maybe add some comments about how the execution test could be run, hbase setup etc.

shawkins:
I can also just make these changes and commit if you want.

kylinsoong:
I have added comments for how to setup HBase and Phoniex to run execution test.

I can't find build/kits/docs/teiid/teiid-releasenotes.html and $mvn -P arqullian-tests clean install run also success, probably I am not update to latest code.

Please feel free to make changes.

I use eclipse 4.4 Luna, I have tried use expand tabs with spaces, but if I do so, all code need format, so I didn't format all code.

shawkins:
Squashed and pulled into master.

PR #414, opened 2015-02-26
TEIID-3121: Changed to package Teiid server as part of Maven build and updated Docker image to use it.

Tom9729:


rareddy:
I want to make sure you did not pull the Teiid dist image from the maven, I only want EAP and Console coming out of it.

Tom9729:
Ramesh,

It should be using the Teiid dist from the build. I changed the jboss-as7/dist.xml to leave another copy of the dist unpacked in the target folder, and the server.xml assembly runs later and pulls this off of the filesystem.

Please let me know if I misunderstood something.

Thanks!

build/assembly/jboss-as7/dist.xml
```diff
@@ -5,6 +5,7 @@
   
   <formats>
     <format>zip</format>
+    <format>dir</format>
```
Tom9729: Leaves Teiid jboss dist unpacked in target folder for use in the server assembly that runs later.

build/assembly/server.xml
```diff
@@ -0,0 +1,34 @@
...
+    </fileSet> 
+    <fileSet>
+        <directory>${project.build.directory}/${dir.jboss-orig}</directory>
+        <outputDirectory>/</outputDirectory>      
+        <excludes>
+          <exclude>**/*.sh</exclude>
+        </excludes>      
+    </fileSet>
+    <fileSet>
+        <directory>${project.build.directory}/teiid-${project.version}-jboss-dist</directory>
```
Tom9729: Pulls in unpacked Teiid jboss dist from filesystem that was created earlier in build.

build/kits/docker/Dockerfile
```diff
@@ -10,33 +10,23 @@ FROM jboss/base-jdk:7
...
+VOLUME ["$JBOSS_HOME/standalone", "$JBOSS_HOME/domain"]
 
 USER jboss
 
 EXPOSE 9990 8080 31000 35432
 
 #Run Teiid Server
-CMD ["/bin/sh", "-c", "/opt/jboss/jboss-eap-6.3/bin/standalone.sh", "-c", "standalone-teiid.xml", "-b", "0.0.0.0", "-bmanagement", "0.0.0.0"]
-
+CMD ["/bin/sh", "-c", "$JBOSS_HOME/bin/standalone.sh -c standalone-teiid.xml -b 0.0.0.0 -bmanagement 0.0.0.0"]
```
Tom9729: The command you want sh to run must be one argument. Without this change, "docker run" would bring up JBoss without changing ports and using the default standalone xml.

rareddy:
Looks good. Only thing I did not understand is the removal of the "release" profile, I remember we had additional tasks in it that are not done during the "dev" profile like source, docs etc. So that during the day to day we not waiting for build to do all the tasks, but only what we are interested in. If you can re-instate that that would be good. Also, can you submit this against 8.10.x branch ? Thanks.

Tom9729:
The release profile still exists, but I removed it from the "build" module because it appeared to be redundant. The "build" module is already conditionally included in the parent. If you do a build without release profile enabled, the "build" module is not included in the reactor. 

See here https://github.com/Tom9729/teiid/blob/TEIID-3121/pom.xml#L158 and here https://github.com/Tom9729/teiid/blob/TEIID-3121/pom.xml#L1515.

I think what would be better would be to always include the "build" module and then use properties to skip "build" plugins (dependency unpack, assembly, retro translator), similar to how Javadoc generation is skipped for dev.

Tom9729:
Fixed; it was removed because of a misunderstanding on my part.

Tom9729:
Anything preventing this from getting merged into master?

rareddy:
no, I waiting on couple more Docker fixes before the push.

shawkins:
Is there anything to be done here to close this pull request?

rareddy:
thought already did this, no pull it.

shawkins:
there's a  lot of conflicts in the dockerfile and the root pom.xml.  Can you have a look?

Tom9729:
I'm not sure what happened with this PR but it looks like it was already merged.

PR #416, opened 2015-03-03
Multiple ddl files in one static VDB

adeelamin:
i did some changes in DDLFileMetadataRepository.java file to add support of multiple ddl files in one static vdb in teiid ,
now multiple ddl files can be add in vdb.xml like
<metadata type="DDL-FILE">
     /ddl/CM_Def.ddl,/ddl/CM_Project.ddl,
     /ddl/CM_Field.ddl,/ddl/Rem.ddl
  </metadata>
 
 
and DDL script files will be place in the DDL folder.
Order of files is depend on sequence which is defined in vdb.xml
 
 
I am splitting the path of files based on the comma delimiter, and then merging the all files in one temp file, and storing that temp file in temp folder,
and then passing the stream of temp file for parsing,
after parsing, temp file will be delete from temp folder

### DIFF
--- a/engine/src/main/java/org/teiid/query/metadata/DDLFileMetadataRepository.java
+++ b/engine/src/main/java/org/teiid/query/metadata/DDLFileMetadataRepository.java
@@ -21,11 +21,17 @@
  */
 package org.teiid.query.metadata;
 
+import java.io.BufferedReader;
+import java.io.BufferedWriter;
+import java.io.File;
+import java.io.FileInputStream;
+import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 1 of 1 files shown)


shawkins:
Moved this to https://issues.jboss.org/browse/TEIID-3372

shawkins:
Thanks for taking a pass at this by the way.  Let us know if the new logic works for you.  You can build from master or there will be an 8.11 alpha1 in the next week.

adeelamin:
shawkins ,

thanks buddy , i wanted to complete this part, but due to some official commitment, didnt get time to complete this .

really appreciating your efforts, i will check this part,

PR #435, opened 2015-03-30
TEIID-3409 PostgreSQLExecutionFactory TranslatorProperty annotation in wrong place

xianliu:
...tGisVersion when using postgresql jdbc translator

when I use postgresql translator of Teiid 8.10.0, it throw when deploying on jboss-as-7.4
the @TranslatorProperty annoation position is not right accoring to code in org.teiid.deployers.TranslatorUtil
getSetter method code
if (method.getName().startsWith("get")) { //$NON-NLS-1$
	setter = "set"+setter.substring(3);//$NON-NLS-1$
}
else if (method.getName().startsWith("is")) { //$NON-NLS-1$
	setter = "set"+setter.substring(2); //$NON-NLS-1$
}
else {
	setter = "set"+method.getName().substring(0,1).toUpperCase()+method.getName().substring(1); //$NON-NLS-1$
}
this annoation can only be annoated on getXXX(), or isXXX() method.

### DIFF
--- a/connectors/translator-jdbc/src/main/java/org/teiid/translator/jdbc/postgresql/PostgreSQLExecutionFactory.java
+++ b/connectors/translator-jdbc/src/main/java/org/teiid/translator/jdbc/postgresql/PostgreSQLExecutionFactory.java
@@ -762,20 +762,20 @@ public class PostgreSQLExecutionFactory extends JDBCExecutionFactory {
     	};
     }
     
-	@TranslatorProperty(display="PostGIS Version", description="The version of the PostGIS extension.",advanced=true)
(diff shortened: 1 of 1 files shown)


xianliu:
@vhalbert @rareddy please have a code & merge it into the repo, it block my usage.

kylinsoong:
@xianliu , thanks for reports, you can discuss the issue with project lead via JIRA in [1] first. Another we usually use JIRA id as comments prefix.

[1] https://issues.jboss.org/browse/TEIID-3409
[2] https://github.com/teiid/teiid/pull/430

xianliu:
ok, got it

PR #436, opened 2015-03-31
TEIID-3393: Add an adminapi interface impl for embedded

kylinsoong:
* Add getAdmin() interface and Admin impl in EmbeddedServer
* Add Embedded AdminAPI unit test code


kylinsoong:
@shawkins 
Any comments of current implementation?

rareddy:
Kylin,  I think you externalize the AdminFactory and AdminImpl out of the EmbeddedServer class, and provide a way for the developer to easily extend, so that they can provide the data source specific methods easily.

Also make sure when you are returning collections do not return the original methods, but put them in separate collection. 

Why netty dependency?

kylinsoong:
Thanks @rareddy .
> externalize the AdminFactory and AdminImpl out of the EmbeddedServer

I would like move it to admin folder, but one I considered is cycle dependency occur, runtime depend on admin, admin depend on runtime, is this fine?
>  netty dependency?

This not added by me

rareddy:
I mean move them into their own classes, not to admin module. Keep em along side the EmbeddedServer.

Sorry, I see my mistake about Netty

kylinsoong:
datasources and resource-adapter subsystem  related  methods and  Deprecated methods not implemented at the moment.

PR #441, opened 2015-04-22
TEIID_3424 enabled support for LT, GT, and NE to be pushed down

vhalbert:


vhalbert:
Can you also back port this to the 62-8.7.x branch?  Thanks.

shawkins:
Did you mean to disable join support?

vhalbert:
Thanks for catching that.

shawkins:
Also DSLSearch.buildQueryFromWhereClause does not handle NOT.  You can specifically enable <. and less than with supportsCompareCriteriaOrderedExclusive with TEIID-3210

vhalbert:
It handles NE (< > ) equality.  Example:   x < > y

Ah, for supportsCompareCriteriaOrderedExclusive  didn't catch the addition post 8.7.   I'll include that with the new jira.

But you're right.  Negation isn't handled when its used like:  Not In (1,2...)  or  Not Like etc.   I'll log a jira for those changes.

PR #443, opened 2015-04-23
Changes for TEIID-3424 and 3453, both related to infinispan-dsl fixes

vhalbert:
Changes for TEIID-3424 and 3453, bot related to infinispan-dsl translator fixes.  And updated cli scripts to add teiid logging context.


shawkins:
Does this mean that the other two pull requests can be closed?

vhalbert:
No, these are different.  The one on master shouldn't be back ported.   I committed these because there are changes back to using getNameInSource in 8.7 (and i added cli changes for product).

PR #444, opened 2015-04-27
TEIID-3431: replace bittronix in the examples with ironjacamar

kylinsoong:
Add EmbeddedHelper Class for supply get TransactionManager and new DataSource functionality


kylinsoong:
@shawkins I added EmbeddedHelper for TransactionManager and replace bittronix, do you think it's okay?

PR #445, opened 2015-04-27
TEIID-3453 and TEIID-3424 to fix infinispan searching 

vhalbert:
TEIID-3453 and TEIID-3424 to fix infinispan searching


shawkins:
Sorry I inadvertently stole your commit.  I squashed the changes, but they were reset.

PR #455, opened 2015-05-07
Updated example to match the local-cache quickstart changes

vhalbert:


vhalbert:
This can also be pushed to master.

PR #462, opened 2015-05-15
TEIID-3475: Add a separate example module for examples

kylinsoong:
### From the functionality aspect, I have adding the following:
* Use a java swing UI list all examples, which the examples can be select and run separately
* Add a Table Renderer utils, which the output will be rendered and formated
* Add transaction manager help class to this module

### Main changes
* move 'build/kits/embedded/examples' to 'examples/kits/embedded/examples'
* move split .java to 'examples/src/main/java'
* combine split run.sh/run.bat to single runExamples.sh/runExamples.bat

### Current embedded dist layout
~~~
[teiid-8.11.0.Beta3-SNAPSHOT-embedded-dist]$ ls
COPYRIGHT.txt  examples  lib  LICENSE-lgpl-2.1.txt  optional  readme.txt  runExamples.bat  runExamples.sh  sources
~~~
Note that:
* runExamples.bat/runExamples.bat  - the script for start java swing UI
* examples - the .vdb, data, config file  which examples depended 
* lib/optional - the jars collect from different module's embedded-dist
* sources - source code of examples


kylinsoong:
@shawkins , do you have any comments about this request?

shawkins:
Not yet.  I haven't had a chance to fully review it.

shawkins:
This looks good.  Also if you want we can work this in a branch.  Here's my thoughts in no particular order.

- set your IDE to replace tabs with spaces
- the example and narayana jars should be in optional not lib
- the example sources should also be under examples
- TeiidExamplesApp has "Embedded dataroleso" - with an extra o at the end
- TableRenderer looks like it needs attribution as which project it came from and may need to retain the original license
- runExamples should probably be called runExamplesApp to clarify that it's a GUI
- there should ideally still be a non-gui way to run the examples and to easily navigate to the java source from the individual source
- Could the examples app capture sys out/err for showing the results of run - that would save looking back at the console
- Modifying the VDB in the example app should probably be removed.  It's not a full editor, does not save changes in between runs, and without an interactive option of what queries to run seems like it won't likely show the user much

kylinsoong:
I added a new updates base on your suggests.

> TableRenderer looks like it needs attribution as which project it came from and may need to retain the original license
The original code come from (http://henplus.sourceforge.net/), but I have added some logic like flush, etc.

> Modifying the VDB in the example app should probably be removed. 
The main I consider is provide  a way for quick view the vdb.

The others all be addressed.

kylinsoong:
@shawkins TableRenderer and ResultRenderer be removed, the graphical launcher UI be move to one of my repository, can you review current updates?

shawkins:
Yes I'll have a look.  Thanks, Kylin.

shawkins:
closing this based upon the direction of the issues

PR #467, opened 2015-05-21
teiid-3151

tejones:
Initial commit of HANA translator.


shawkins:
Is the concat modifier needed?  It looks like https://help.sap.com/saphelp_hanaplatform/helpdata/en/20/a380977519101494ceddd944e87527/content.htm they follow ansi where if either value is null the result is null

PR #468, opened 2015-05-29
TEIID-3491 Added REGEXP_REPLACE.

Tom9729:


engine/src/test/java/org/teiid/query/processor/TestGeometry.java
```diff
@@ -192,19 +192,6 @@ public void testAsKmlException() throws Exception {
         assertEval("ST_AsKML(ST_GeomFromText('POLYGON((0 0,0 1,1 1,1 0,0 0))'))", null);
     }
     
-    private void assertEval(String expr, String result)
```
Tom9729: Moved this into ```TestFunctionResolving ``` so I could use it to test regex expressions.

PR #476, opened 2015-06-18
TEIID-3541: remove embedded kit

kylinsoong:
* remove 'embedded-dist.xml'
* update assembly-plugin descriptors, remove 'embedded-dist.xml' descriptor


shawkins:
Before we merge this commit we'll have to normalize the embedded dependencies, we currently in the kitting have a mixture of:
just bringing in the teiid artifacts
using transitive dependencies (possibly with excludes)
using explicit dependencies

switching to explicit dependencies is preferred, but will possibly require updating the AS kitting

kylinsoong:
@shawkins , do we need keep this PR? I can't create multiple PR for Vertica translator right now, so I wonder if I can close this, due to there are some other works need to do, like updating the AS kitting?

shawkins:
Yes, if we need to close this to bring in Vertica, then let's do that for now.

PR #512, opened 2015-08-17
TEIID-3627 disabling CompareCriteriaOrdered until JDG gets a fix

vhalbert:


vhalbert:
This should committed to the 62-8.7.x and 8.11.x (unless 8.12 is now the target for DV 6.3)

PR #514, opened 2015-08-18
TEIID-3521: Use Swagger UI add a api-doc for VDB Rest Service

kylinsoong:


kylinsoong:
This pull not complete, but any comments/advice is appreciate.

jboss-integration/src/main/java/org/teiid/jboss/rest/RestASMBasedWebArchiveBuilder.java
```diff
@@ -506,14 +602,29 @@ private void buildQueryProcedure(String vdbName, int vdbVersion, String modelNam
 			}
 			{
 			av0 = mv.visitAnnotation("Ljavax/ws/rs/Path;", true);
-			av0.visit("value", "/query");
+			av0.visit("value", modelName);
```
rareddy: Why change to modelname? This is a generic service, that one can execute, so need to have a generic name

pom.xml
```diff
@@ -1545,6 +1557,7 @@
 		<module>hibernate-dialect</module>
 		<module>eclipselink-platform</module>
 		<module>jboss-integration</module>
+		<module>swagger-integration</module>
```
rareddy: There is no need to for separate "Swagger-integration" module, move this into test-integration-common

rareddy: How is swagger-integration used? If it is for testing this needs to be in "test-integration-common", you need the bootstrap servlet then this needs to part of the byte code generation code.

jboss-integration/src/main/java/org/teiid/jboss/rest/RestASMBasedWebArchiveBuilder.java
kylinsoong: This is a miss type, will roll back

pom.xml
kylinsoong: io.swagger's dependencies conflicts with jboss-integration's dependencies(jackson, etc), this the reason I use a separate module at the moment.

 it's not for test, current it use a servlet to bootstrap the Swagger(scan the anootations, form the swagger.json to Swagger UI JS)

rareddy: Is it "com.fasterxml.jackson"? what exactly are conflicts?

jboss-integration/src/test/java/org/teiid/jboss/rest/TestRestWebArchiveBuilder.java
```diff
@@ -64,10 +71,59 @@ public void testBuildArchive() throws Exception {
 		ArrayList<String> files = new ArrayList<String>();
 		files.add("WEB-INF/web.xml");
 		files.add("WEB-INF/jboss-web.xml");
+		files.add("WEB-INF/jboss-deployment-structure.xml");
 		files.add("WEB-INF/classes/org/teiid/jboss/rest/View.class");
 		files.add("WEB-INF/classes/org/teiid/jboss/rest/TeiidRestApplication.class");
 		files.add("META-INF/MANIFEST.MF");
 		
+		files.add("api-doc.html");
```
rareddy: Also do not include all the required "Static" files one by one to the archive, just add them from pre-packed directory of swagger stuff, that way when we need to update swagger version, it is just updating whatever that new directory

kylinsoong:
@rareddy can you review and consider merge into teiid repo? I have done some initial test, it worked fine, probably we can do some work like polish the UI, etc in future with QE/community user's request.

build/kits/jboss-as7/modules/io/swagger/main/module.xml
```diff
@@ -0,0 +1,37 @@
...
+        <property name="jboss.api" value="private"/>
+    </properties>
+    
+    <resources>
+        <resource-root path="swagger-annotations-${version.swagger-jaxrs}.jar" />
+        <resource-root path="swagger-core-${version.swagger-jaxrs}.jar" />
+        <resource-root path="swagger-jaxrs-${version.swagger-jaxrs}.jar" />
+        <resource-root path="swagger-models-${version.swagger-jaxrs}.jar" />
+        <resource-root path="jackson-dataformat-xml-${version.com.fasterxml.jackson.swagger}.jar" />
+        <resource-root path="jackson-dataformat-yaml-${version.com.fasterxml.jackson.swagger}.jar" />
```
rareddy: Shouldn't all the "jackson" dependencies need to their own module, if there is already module, we need to create a new version of that module.

jboss-integration/src/main/java/org/teiid/jboss/rest/RestASMBasedWebArchiveBuilder.java
```diff
@@ -117,7 +119,280 @@ else if (securityType.equalsIgnoreCase("httpbasic")) {
 		return byteStream.toByteArray();
 	}
 
-	private void writeEntry(String name, ZipOutputStream out, byte[] contents) throws IOException {
+    private void writeSwagger(ZipOutputStream out, Properties props, VDBMetaData vdb) throws IOException {
```
rareddy: This is not what I meant, I want you to add the "swagger" directory with its contents to the build somewhere, then just add the directory, that way when time to upgrade the swagger we can just replace that with new contents. We do NOT want to add one file by one file to the zip

jboss-integration/src/main/java/org/teiid/jboss/rest/RestASMBasedWebArchiveBuilder.java
```diff
@@ -117,7 +119,280 @@ else if (securityType.equalsIgnoreCase("httpbasic")) {
...
+            mv.visitEnd();
+        }
+        
+        cw.visitEnd();
+        
+        return cw.toByteArray();
+           
+    }
+	
+	protected byte[] getBootstrapServletClass(String vdbName, String desc, String version, String[] schamas, String baseUrl, String packages, Boolean scan) {
```
rareddy: Is there any *dynamic* content in here? if there not any, we need to add this as  .java class, and then use that.

kylinsoong: ApiOriginFilter didn't contain dynamic code,  BootstrapServlet contain dynamic code, what VDB info(version, name) need to pass to BootstrapServlet's init() and doPost() method.

If we use .java rather than byte base code, how to place the .java code? My original thoughts is use a separate module which contain all 
* ApiOriginFilter.java
* BootstrapServlet.java
* swagger dist
it be built to a jar, merge to war when vdb deploying. What's your idea for where to place the .java?

jboss-integration/src/main/java/org/teiid/jboss/rest/RestASMBasedWebArchiveBuilder.java
kylinsoong: @rareddy 
> contents to the build somewhere,
replaceTemplates for all swagger static file is not necessary, we can't add all swagger directory via build, because the war generated dynamicly in deploying vdb, but we can recurse the swagger directory add each file as a zipEntry.

build/kits/jboss-as7/modules/io/swagger/main/module.xml
kylinsoong: @rareddy 
Yes, I will split jackson-dataformat, jackson-datatype, jackson-module-jaxb to it's own module.

jboss-integration/src/main/java/org/teiid/jboss/rest/RestASMBasedWebArchiveBuilder.java
rareddy: Put all the code in .java, and then have "Myclass extends BootstrapServlet" with the dynamic content class in Byte code

kylinsoong:
@rareddy 
I would like add a summary as below:

## Summary
### split swagger module
swagger module be split into the following sub-module, add dependencies base on dependency tree:
* jackson.dataformat.xml
* jackson.dataformat.yaml
* jackson.datatype
* jackson.jaxrs
* org.reflections
* com.google.code.findbugs

### use .java replace byte based code
ApiOriginFilter.java and BootstrapServlet.java be added, only init() methods be overwrite via byte code for some dynamic info.

### use .zip replace the static folder.
swagger-ui-2.1.1.zip be added, writeDirectoryEntry used to write all ZipEntry exist in zip to ZipOutPutStream.
I spend time to merge bytes directly without ZipOutPutStream, but not work, so I use about method.

rareddy:
Looks like there are conflicts, please fix and submit a clean pull

kylinsoong:
Change will submit to new pull request

PR #518, opened 2015-08-31
TEIID-3667 LDAPSyncQueryExecution.createSearchContext() hides a root …

hokuda:
…cause when an error occurs


shawkins:
Ended up just removing the log error statement instead.

PR #519, opened 2015-08-31
Added type mapping for BigDecimal to Decimal

dkrapohl:
Added missing type mapping for Teiid bigdecimal to Impala Decimal data type


shawkins:
It looks like support was added for Imapla 2.1.  So we would need to conditionally add support for decimal or update the docs to say change our Impala support which currently starts at 1.2.1.

shawkins:
Thanks, I rolled this into https://issues.jboss.org/browse/TEIID-3700

PR #520, opened 2015-09-03
Changed for product build only to produce javadocs

vhalbert:
Changed for product build only, so the javadocs can be produced at the same time as the -Prelease is done, in one step.

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -190,25 +190,6 @@
           This profile is activated manually, as in "mvn ... -P release ..."
 		  -->
       <id>release</id>
-      <properties>
-        <maven.javadoc.skip>true</maven.javadoc.skip>
-      </properties>
-      <modules>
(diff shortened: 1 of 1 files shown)


shawkins:
Is there a reason not to use the existing releasewithjavadoc profile?

vhalbert:
in the build/pom.xml is looking for release profile, and want to do it all in one step.   Could have added another profile there.  But talking with prod guys, it just sound simpler to just change the root pom.

rareddy:
so basically you want to use the profile name "release". "releasewithjavadoc" is exactly what you looking for

shawkins:
He's also saying that the build pom is only activating under the "release" profile as well.  The least painful thing to do would probably be to use this change and update the release docs to add the skip docs property to the command line.

PR #524, opened 2015-09-14
TEIID-3690: add a ExceptionMapper for generated rest war

kylinsoong:
The ExceptionMapper will return error as a xml:
~~~
<error>
  <code>ERROR</code>
  <message>message</message>
  <details>stacktrace</details>
</error>
~~~

### DIFF
--- a/build/kits/jboss-as7/modules/org/jboss/teiid/main/module.xml
+++ b/build/kits/jboss-as7/modules/org/jboss/teiid/main/module.xml
@@ -66,8 +66,9 @@
         <module name="org.jboss.as.controller-client"/> <!-- Admin -->
         <module name="org.jboss.as.controller"/> <!-- Admin -->
         <module name="io.swagger"/>
-		<module name="javax.servlet.api"/>
-		<module name="org.apache.commons.io"/>
+	<module name="javax.servlet.api"/>
+	<module name="org.apache.commons.io"/>
+	<module name="org.jboss.resteasy.resteasy-jaxb-provider" />
         
     </dependencies>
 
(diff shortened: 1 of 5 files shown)


rareddy:
See if you can capture the right "http status" code, rather than "ERROR" string as code.

kylinsoong:
@rareddy current eap 6.4 can not capture "http status", but in new version wildfly, it can capture "http status",I have comment out these code:
~~~
// if(e instanceof NotFoundException){
// code = "404";
// } else if(e instanceof InternalServerErrorException){
// code = "500";
// }
~~~
Current code can return

kylinsoong:
As document [2] eap 6.4 use resteasy 2.3.10.Final, base on document [1], it should work, let me re-test it tomorrow.

[1] https://docs.jboss.org/resteasy/docs/2.2.1.GA/userguide/html/ExceptionHandling.html
[2] https://access.redhat.com/articles/112673

rareddy:
The links you showed tells that if you want to use their defined Exception messages you need certain version. I am not sure what kind of exceptions we used the REST wars, but what was wrong with the above commented code?

kylinsoong:
404 can not be captured via ExceptionMapper(I think should be a issue, [1]), this is the reason made be confused yesterday.

From Teiid side, I think commented code is correct, so add another commit, comment out it.

[1] https://issues.jboss.org/browse/RESTEASY-1229

rareddy:
What about JSON payload? If ACCEPTS header is "application/json" then json payload needs to be sent

kylinsoong:
@rareddy, can you review the current pull, do you have any other advice?

PR #538, opened 2015-10-08
TEIID-3520: Add Swagger Metadata support in the WS Translator

kylinsoong:


kylinsoong:
@rareddy Can you review current pull, I have tested both embedded and client/server mode, it all works fine.

rareddy:
Although this is good start, this not complete. I will provide more feedback on the JIRA

PR #541, opened 2015-10-20
Coherence related changed, plus fix for handling limit option and visitor consolidation

vhalbert:
Can you also push these changed to the 8.12.x branch?


shawkins:
Yes, after I review for master, I'll pull to 8.12.x as well.

PR #547, opened 2015-11-05
TEIID-3520: Add Swagger Metadata support in the WS Translator

kylinsoong:
The main change including:
* change `api` module, add getSwaggerURL() method in WSConnection
* change `connector-ws` module, update WSConnectionImpl and WSManagedConnectionFactory for add get swagger URL
* update `pom.xml`, add swagger-parser dependency
* update build assembly/kits
* add `translator-swagger` module for swagger translate support, 

The main `translator-swagger` did including:
* SwaggerMetadataProcessor use to parse swagger json file load each opertaion as a procedure, opertaionID as procedure name
* SwaggerProcedureExecution use to execute procedure, base on ws resource adapter
* PUT, POST, GET, DELETE be supported
* produces and consumes in Metadata swagger json file used to validate api request and response  context


connectors/translator-swagger/src/main/java/org/teiid/translator/swagger/SwaggerMetadataProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+            procedure.setProperty(HTTP_HOST, httpHost); 
+            procedure.setProperty(PROCEDURE_PRODUCES, produces);
+            procedure.setProperty(PROCEDURE_CONSUMES, consumes);
+            procedure.setAnnotation(annotation);
+            
+            mf.addProcedureParameter("result", TypeFacility.RUNTIME_NAMES.BLOB, Type.ReturnValue, procedure); //$NON-NLS-1$ 
+            
+            addProcedureParameter(mf, procedure, operation);
+            
+            ProcedureParameter param = mf.addProcedureParameter("headers", TypeFacility.RUNTIME_NAMES.CLOB, Type.In, procedure); //$NON-NLS-1$
```
rareddy: Are "headers" used as input parameters in Swagger to define the behavior of the call? I mean is user need to enter their own specific data for headers?

If yes, then we need to collect each of them as input parameter. if no, we need to collect what value needs to be sent for which header and keep it properties and send it at execution time. Having one enter a clob is not good usability wise.

connectors/translator-swagger/src/main/java/org/teiid/translator/swagger/SwaggerMetadataProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+            param.setAnnotation("Headers to send"); //$NON-NLS-1$
+            param.setNullType(NullType.Nullable);
+            
+            param = mf.addProcedureParameter("contentType", TypeFacility.RUNTIME_NAMES.STRING, Type.Out, procedure); //$NON-NLS-1$
```
rareddy: This should not be needed, more on this coming below. If the service produces both XML/JSON type payloads, pick one like JSON to parse. Add a translator property, what is user's preference to read the document as.

connectors/translator-swagger/src/main/java/org/teiid/translator/swagger/SwaggerMetadataProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+            } else if(parameter instanceof QueryParameter) {
+                QueryParameter p  = (QueryParameter) parameter;
+                String type = p.getType();
+                dataType = DataTypeManager.getDataTypeName(TypeFacility.getDataTypeClass(type));
+                isPathParam = FALSE_STRING; //$NON-NLS-1$
+            }
+            ProcedureParameter param = mf.addProcedureParameter(name, dataType, Type.In, procedure); 
+            param.setProperty(IS_PATH_PARAMETER, isPathParam); 
+            param.setNullType(NullType.No_Nulls);
+            param.setDefaultValue(FALSE_STRING); //$NON-NLS-1$
```
rareddy: ? why false string? if you got nothing do not define it

connectors/translator-swagger/src/main/java/org/teiid/translator/swagger/SwaggerMetadataProcessor.java
```diff
@@ -0,0 +1,286 @@
...
+            procedure.setNameInSource(procName);
+            procedure.setProperty(HTTP_ACTION, action);
+            procedure.setProperty(HTTP_API_PATH, key);
+            procedure.setProperty(BASE_URL, baseUrl);
+            procedure.setProperty(HTTP_HOST, httpHost); 
+            procedure.setProperty(PROCEDURE_PRODUCES, produces);
+            procedure.setProperty(PROCEDURE_CONSUMES, consumes);
+            procedure.setAnnotation(annotation);
+            
+            mf.addProcedureParameter("result", TypeFacility.RUNTIME_NAMES.BLOB, Type.ReturnValue, procedure); //$NON-NLS-1$
```
rareddy: We can NOT have blob as return, that is what we did before when we did not knew what is being returned. The value of Swagger is it defines what is being returned. You have "ref" element pointing to a data structure, that needs to be return of this procedure. If the blob is the return the user still needs to write parsing logic using XMLTABLE in a view transformation. There we are not providing much value, other than just providing SQL way execute the REST API.

This is what I was saying before, you have two choices here. One to create virtual procedure, that uses the XMLTABLE semantics to parse the payload into a tabular structure, or write parser for JSON and XML. Since Swagger tell how these payloads are formatted it should be easy enough to do either. If the payload type is unknown or not a JSON/XML then use the blob.

connectors/translator-swagger/src/main/java/org/teiid/translator/swagger/SwaggerMetadataProcessor.java
kylinsoong: "Content-Type", "Accept" is needed which be passed as jsonObject, like ws-translator, then in runtime, parse jsonObject, add as headers
~~~
jsonObject('application/json' as Content-Type, 'application/json' as Accept),  jsonArray('gzip', 'deflate') as \"Accept-Encoding\"))
~~~
can we do like this?

kylinsoong:
@rareddy , I have done some change:
* headers IN parameter used to keep "Content-Type", "Accept" and security related properties, these will be add as httpHeaders in execute.
* I am not use olingo JsonDeserializer(I don't want add more dependencies), instead write a JsonDeserializer, if return Content-Type is application/json, use it convert Inputstream to json string
* The others you referred in above also be handled.

What's your think of Security handle?
I am not detailed investigate the spec, but from swagger api, there are 3 kinds of security mode(HttpBasicAuth, ApiKeyAuth, OAuth) can be handle, there also examples for how to add to headerParams.

Appreciate for any advices/further suggestion, thanks for your time.:)

rareddy:
- It is not clear to me how you are deserilizing the return values. You show a primitive and collection, what about a object like "Customer"? Why they are designed as output parameters?
- You can NOT use Olingo, their de-serializers are specific to their entity model. Yes, you to write your own.
- Will handle security at the end, once we have all others correct.

kylinsoong:
The 2 methods for different ContentType, application/json and application/xml correspondingly:
* if return ContentType is `application/json`, convert stream to json string:
~~~
import com.fasterxml.jackson.databind.ObjectMapper;
ObjectMapper mapper = new ObjectMapper();
Object value = mapper.readValue(input, Object.class);
String jsonString = mapper.writeValueAsString(value);
~~~
* if return ContentType is `application/xml`, convert stream to xml string:
~~~
StringWriter output = new StringWriter();
IOUtils.copy(input, output);
String xmlString = output.toString();
~~~
* If ContentType not json/xml, return Blob which wrap the stream.

rareddy:
No the above is not right, I did not suggest to convert to JSON or XML to String. That is just serializing one format to another, that is not right.

What I am saying the Swagger API defines the return object for all the calls like getCustomer() returns a JSON, that represents "Customer" object. Which may have properties like

Customer {
   String name;
   Date dob;
   String SSN;
} 

etc, now when model the above in the Swagger translator, the Procedure getCustomer should be like

CREATE PROCEDURE GetCustomer(String id) RETUNS TABLE (name string, dob date, ssn string)

and wen JSON and XML is returned from execution of the REST call, then you need to parse it into the "Customer" object and return the tabular results.

kylinsoong:
use a lot of odata4 parse document logic, still not finished.

rareddy:
Please do NOT use OData parsing logic, the payload formats are NOT same. OData document parsing logic comes from libraries like OReva or Olingo

kylinsoong:
What the key principles I use to convert json return to matrix is: Only if a rest api return Object or a list of Object, convert to matrix, the Object's properties be convert to columnName, the Object's Collection properties be treat as a Object, 

for example:
~~~
Customer{
String a;
Integer b;
Product c;
List<Product> d;
Set<Product> e;
}
Product{
String p1;
Integer p2;
Sales p3;
}
Sales {
String s1;
Integer s2;
}
~~~

The Customer Object will be convert to a table 

~~~
Customer(
    a, string,
    b, integer,
    c/p1, string,
    c/p2, integer,
    c/p3/s1 string,
    c/p3/s2 integer,
    d, object,
    e, object
)
~~~

kylinsoong:
@rareddy can you review the current code, I have tested on  Customer and PerStore, it works.

connectors/translator-swagger/src/main/java/org/teiid/translator/swagger/SwaggerProcedureExecution.java
```diff
@@ -0,0 +1,546 @@
...
+    private Object getReturnParameter(Procedure procedure) {
+        for (ProcedureParameter pp : procedure.getParameters()){
+            if (pp.getType() == ProcedureParameter.Type.ReturnValue) {
+                return pp;
+            }
+        } 
+        return null;
+    }
+
+    private void applyAuthParams(Procedure procedure, Map<String, List<String>> httpHeaders) throws TranslatorException {
```
rareddy: We can not have any Auth specific code in the translator. It needs be at resource-adapter level. You can have metadata on the translator that it needs to use certain Authentication but not enforce at this layer. So remove all auth specific code.

kylinsoong:
@rareddy 
I have found that it's not possible to open multiple pull requests to a repo, so can we merge and close it, the future change we can do it in new JIRA and new pull.

kylinsoong:
@rareddy Have you planned to re-use these code? since I realized you are work on swagger now.

rareddy:
Yes, I started with your copy, and modifying that.

kylinsoong:
Closed due to diff has created

PR #555, opened 2015-11-23
JBDVPL-270: Use declared dependencies in Teiid assembly POMs instead …

pkocandr:
…of reactor contents

This is the first step which declares needed dependencies in assembly
project which helps us to produce product Maven repository. Also adds
missing dependency management that is needed for declared dependencies.


pkocandr:
I was not able to check if everything is resolvable and set correctly because of missing org.jboss.teiid.connectors.google:gdata-spreadsheet:jar:3.0 which prevented me from building org.jboss.teiid.connectors:google-api so only manual checks were performed.

rareddy:
1) This is not a product branch, this is community. 
2) The usage of "ModuleSet" over "dependencySet" is intentional choice to keep the dependencies of each modules defined in their modules, not in a common project like "build"

Please create a community issue before submitting any pull requests. Rejecting for now.

PR #559, opened 2015-12-02
Upgraded to use community Infinispan 8.x

vhalbert:


rareddy:
Why is change for the "antlr4" that will not work with Olingo

PR #560, opened 2015-12-03
Upgraded to use community infinispan 8.x

vhalbert:
upgraded to community infinispan 8.x.  Also added specific antlr dependency, that infinispan needs, to the infinispan pom's.


rareddy:
Just want to make sure this is the same version as in the WildFly or you are going with your own version  for translator?

vhalbert:
This is the antlr version in the infinispan downloaded kit.

rareddy:
i mean the Infinispan version?

vhalbert:
That's:   <version.org.antlr.infinispan>3.4</version.org.antlr.infinispan>  in the root pom.

rareddy:
Van, you talking about Antlr, I am asking about Infinispan version. What I see from WildFly distribution that it has Infinispan "7.2.3.Final", you changed that to "8.0.1.Final" for the translator, that means we need to supply those JAR files now in Teiid's distribution. Typically *if* the modules are already in the host environment, we try to match our implementation to those versions in WildFly to avoid supplying the jar files and possibly be in tandem with the product version. If we start stray away from that, then the burden is on us to support a different version (could be difference in configuration in standalone.xml etc), which IMO not good. So, unless you got *solid* reason to jump ahead in Infinispan version, I do not think this is good.

vhalbert:
I don't think we want to support the infinispan within wildly.   This is the same way its done in jboss as 7, in which the internal Infinispan isn't supported as a container for user data.

rareddy:
Why? The reason before was that Infinispan moved on in community and JDG was left on older branch where we could not advance it for product aspirations. This is no longer the issue. IMO, we should correct this.

vhalbert:
Actually, its our requirement that we cannot use the internal infinispan for customer data.    Speak to steve J. about this.

shawkins:
To clarify I think you mean our product requirement is to use productized JDG.  It doesn't really make sense to say that this is the community requirement as the infinispan bits are generally available and otherwise promoted for use.

> that means we need to supply those JAR files now in Teiid's distribution

Not necessarily.  We can choose to align ourselves with the product requirement and simply increase the burden on the install or provide a larger EAP alpha + JDG alpha kit.

rareddy:
AFAIK, WildFly 9 is community only. If Infinispan translators are product specific, then we need to look  into how to cleanly separate them from community concerns.

PR #581, opened 2016-01-22
TEIID-3922: translator-accumulo kits added org.jboss.teiid dependency

jstastny-cz:


shawkins:
For product changes we'll target the 63-8.12.x branch

PR #587, opened 2016-01-22
Alignment with new ipbom and updating version

vhalbert:


vhalbert:
Can you commit this and then we can make any adjustments to the version after that?

PR #588, opened 2016-01-23
TEIID-3725 and TEIID-3559 refactoring and adding support for materialization

vhalbert:
TEIID-3725 and TEIID-3559 this includes refactoring of the remote-cache logic to ue translator-object as its base, and changing remote-cache to support materialization


vhalbert:
please also commit this also to the 63_8.12.x branch.

shawkins:
This pulled cleanly into 6.3 8.12.x, but it will take me a bit to get into master.  There the object translator module.xml does say that it should have the user module added, is this correct?

vhalbert:
the user module is set to optional=true, but the user still has add that in order for the correct dependency to be there.

shawkins:
is it possible to use an override property instead?  it generally isn't good to have users modify module.xml files that in the dv base

vhalbert:
Us and JDG have directions for updating module.xml to add the pojo to the class path.

vhalbert:
And its been this way since 6.0 (teiid 8.4.x).

shawkins:
That doesn't mean it should be left that way as that's file that could be affected by patching and forces all usage to share the same file (there is no vdb isolation, and conflicting classes would require a new translator module).

PR #593, opened 2016-01-26
Updates for product build

vhalbert:
Removed SNAPSHOT from version, fixed poms for product build, and had to turn off TestODataIntegration test for now as it will not work due to changes, and fixed the jboss-as kit to align for product


shawkins:
What is your issue with TestODataIntegration?  With these changes the integration tests won't run for me at all since there is no standalone configuration.

vhalbert:
There must have been changes to the class it extends (BaseResourceTest), because the @BeforeClass wouldn't compile.

vhalbert:
ah, for the standalone, we can add that back.    I removed it thinking it was only used for the jboss-as kit, and for which it is eventually removed and not available for use in DV.

shawkins:
Can it be added back to the kit, or should there be something special just for the integration tests?

vhalbert:
it can be added back, not an issue.

shawkins:
The next things that I'm seeing is that the olingo-patches was removed (but is still needed) and it looks like the version.com.fasterxml.jackson.swagger is needed.

vhalbert:
removed?? i see the sub-project, but I had to add the module to the module list in the root pom.xml in order to get it built before the connectors.

shawkins:
removed from the kitting along with teiid-olingo-common.  But those were from the latest changes to the branch so it could have been inadvertant.

vhalbert:
I bet I missed those when I applied the changes from my branch.

shawkins:
Have a look at the next commit to see if there is anything you want to refine from there.

PR #594, opened 2016-01-28
Push  recommended change upstream

vhalbert:
Can this also be backported to 63-8.12.x?


shawkins:
I'll pull 5dc2bf6.  It doesn't look like the coherence commit is needed.

PR #600, opened 2016-02-03
Prevent potential NPE in LDAPManagedConnectionFactory

asmigala:
Could occur when calling hashCode due to unitialized
ldapTxnTimeoutInMillis


shawkins:
Thanks for seeing this.  I associated the commit with the original issue - TEIID-1957 which required us to use wrapper types and not primitives on the connection factory properties.

PR #603, opened 2016-02-04
added command to add AUDIT_LOG to standard install, …

vhalbert:
added command to add AUDIT_LOG to standard install, so that all following installation scripts assume COMMAND_LOG and AUDIT_LOG are configured, changed to WARN level, like the default of COMMAND_LOG is set on initial install


shawkins:
Just to double check, you know that at a warn level you won't see any messages correct?

vhalbert:
dang, thought it was just limiting the detail on the command log, like the user query details, not the source query details.   If thats the case I'll close this PR.

PR #604, opened 2016-02-04
changed standalone and domain to install AUDIT_LOG 

vhalbert:
changed standalone and domain to install AUDIT_LOG so that any followup cli script assumes its there and will therefore have a remove command.


vhalbert:
Can I get this committed?

shawkins:
This also needs rolled forward correct?  

Does the hana change need to be managed separately - will it be needed in 8.7.x as well?

vhalbert:
Not needed in 8.7.x, its already out the door.   As for rolled forward, I don't think it hurts anything, and it would eliminate the need to make this change again.

shawkins:
I'm also seeing that the jpa2, simpledb, and ucanaccess translators are not in the cli scripts.  Those are needed as well correct?

vhalbert:
I've got another set of changes coming, so I'll add those back then.

PR #605, opened 2016-02-09
updating local cache to support materialization

vhalbert:


shawkins:
Does this go to master as well?

vhalbert:
ah, yes, please, these go along with the last set of jdg related changes.

shawkins:
Which set of changes are you referring to?

PR #607, opened 2016-02-12
TEIID-3974 use tinyint when calling min/max on Vertica

asmigala:


shawkins:
Thanks for the fix Andrej

PR #614, opened 2016-02-18
removed version from the names of artifacts, add teiid-hibernate-dialect to the kit, and udpated cli scripts to deploy artifacts with no version

vhalbert:


shawkins:
Pulled the last commit.

PR #618, opened 2016-02-24
Adding back the spring framework jars needed for soap

vhalbert:


shawkins:
Hasn't the kitting been this way since https://issues.jboss.org/browse/TEIID-3185 - https://github.com/teiid/teiid/blob/e526c6641422f85b7dd6112909b3b2c78767ecaa/connectors/connector-ws/kits/jboss-as7-dist.xml

Has there been paths that haven't been working since 8.9?  Do we need an integration test here?

vhalbert:
This issue was only found from the use of SOAP generation from Designer.   My guess is not many people have tried this feature in the community.

rareddy:
Are you sure you do not need the versions at the end of those jars?

vhalbert:
i think ur right

vhalbert:
I verified the jar's where getting put there, and

PR #621, opened 2016-02-25
Adding back in the spring jars to support SOAP services

vhalbert:


rareddy:
Can you also do the same for master?

vhalbert:
i was having issues building master, so once that is resolved I can commit it.

PR #624, opened 2016-03-01
TEIID-4007 - fixed extraction of name of the model if context root is

jdurani:
Fixed extraction of name of the model if context root is changed.


shawkins:
Merged with some additional refinements.

PR #625, opened 2016-03-03
TEIID-2820 - added couchbase resource adapter and translator scaffolding 8.12.x

jmarley:
This pull request adds the couchbase translator. I currently have the resource adapter working, but haven't had a chance to start translator.


shawkins:
It's still being determined if this should go into 8.12.5.   In the meantime, would it be possible for you to squash these into a single commit against master?

shawkins:
After looking through this a little more.  I don't think we are quite ready for the sandbox area yet.  It would probably be best to spin this off with its own repository location and then pull it in when it is functionally complete.

PR #645, opened 2016-03-21
TEIID-4094 refactoring teiid-jdbc.jar to be in its own project

vhalbert:


shawkins:
What is the purpose of the dependency-reduced-pom.xml?

PR #661, opened 2016-04-05
TEIID-3480: use matViewStatus as a unique endpoint for getting both i…

kylinsoong:
The change will make sure 
~~~
EXEC SYSADMIN.matViewStatus('Schema', 'MatView')
~~~
will return status for both internal and external.


kylinsoong:
So far SYSADMIN.matViewStatus can used to query status for both internal and external, In my test, there are 2 VDB:
* PortfolioInterMaterialize - which demonstrate internal mat, stockPricesInterMatView exist in StocksMatModel
* PortfolioMaterialize - which demonstrate external mat, stockPricesMatView exist in StocksMatModel
Then query status via CLI:
~~~
[standalone@localhost:9990 /] /subsystem=teiid:execute-query(sql-query="EXEC SYSADMIN.matViewStatus('StocksMatModel', 'stockPricesInterMatView')", vdb-name=PortfolioInterMaterialize, vdb-version=1,timeout-in-milli=1000)
{
    "outcome" => "success",
    "result" => [{
        "TargetName" => "#MAT_STOCKSMATMODEL.STOCKPRICESINTERMATVIEW",
        "Valid" => true,
        "LoadState" => "LOADED",
        "Updated" => "2016-04-06 15:23:05.281",
        "Cardinality" => 18L,
        "LoadNumber" => -1L,
        "OnErrorAction" => "THROW_EXCEPTION"
    }]
}
[standalone@localhost:9990 /] /subsystem=teiid:execute-query(sql-query="EXEC SYSADMIN.matViewStatus('StocksMatModel', 'stockPricesMatView')", vdb-name=PortfolioMaterialize, vdb-version=1,timeout-in-milli=1000)
{
    "outcome" => "success",
    "result" => [{
        "TargetSchemaName" => "Accounts",
        "TargetName" => "H2_STOCK_MAT",
        "Valid" => true,
        "LoadState" => "LOADED",
        "Updated" => "2016-04-06 15:23:06.185",
        "Cardinality" => 18L,
        "LoadNumber" => 2L,
        "OnErrorAction" => "THROW_EXCEPTION"
    }]
}
~~~
Now, I plan to update teiid-web-console to let ui consistent.
@shawkins can you review this pull request, any idea?

shawkins:
SYSADMIN.matViewsStatus() won't work as there isn't a single status table for the whole vdb.  We could look at eventually providing status over each status table.  For now the changes to just account for internal in the existing matviewstatus will work.

shawkins:
I ended up squashing this and adding a unit test.  Thanks Kylin.

kylinsoong:
`SYSADMIN.matViewStatus('Schema', 'MatView')` has precondition, which means we have to know the Model name, and MatView name. But in most scenario we need view the status by VDB.

IMO, status table in external also should be unique, like internal, all status of external mat view should persist to this unique status table, have create TEIID-4121

kylinsoong:
I would keep `matViewsStatus()` in case in future we may need this
~~~
CREATE VIRTUAL PROCEDURE matViewsStatus() RETURNS TABLE (VDBName varchar(255), SchemaName varchar(255), Name varchar(255), TargetSchemaName varchar(255), TargetName varchar(255), Valid boolean, LoadState varchar(255), Updated timestamp, Cardinality long) AS
BEGIN
    DECLARE string vdbName = (SELECT Name FROM VirtualDatabases);
    DECLARE integer vdbVersion = (SELECT convert(Version, integer) FROM VirtualDatabases);

    /*
      In current design, status table on the physical database, and used in any one of the source models
    */
    DECLARE string statusTable = (SELECT "Value" from SYS.Properties WHERE Name = '{http://www.teiid.org/ext/relational/2012}MATVIEW_STATUS_TABLE' LIMIT 1);

    IF (statusTable IS NULL)
    BEGIN
        EXECUTE IMMEDIATE 'SELECT VDBName, SchemaName, Name, TargetSchemaName, TargetName, Valid, LoadState, Updated, Cardinality FROM MatViews' AS VDBName string, SchemaName string, Name string, TargetSchemaName string, TargetName string, Valid boolean, LoadState string, Updated timestamp, Cardinality long USING vdbName = VARIABLES.vdbName;
    END ELSE
    BEGIN
        EXECUTE IMMEDIATE 'SELECT VDBName, SchemaName, Name, TargetSchemaName, TargetName, Valid, LoadState, Updated, Cardinality FROM ' || VARIABLES.statusTable AS VDBName string, SchemaName string, Name string, TargetSchemaName string, TargetName string, Valid boolean, LoadState string, Updated timestamp, Cardinality long USING vdbName = VARIABLES.vdbName, vdbVersion = VARIABLES.vdbVersion;
    END
END
~~~

shawkins:
In the existing form it will not work, but it can be attached to https://issues.jboss.org/browse/TEIID-4121 and used if needed. There will need to be some more discussion on that issue.

PR #665, opened 2016-04-12
TEIID-3725 updating release notes

vhalbert:


shawkins:
I went ahead and pulled this as I'm assuming you had reasons for the formatting changes - but it does seem odd using non-enclosing li tags (it causes my editor to complain about all of the p tags that then are children of a ul element).  So at this time I'm just going to pull the highlight items and not the whole change to master.

PR #666, opened 2016-04-13
TEIID-4112 Fix recursive CTE pushdown with Oracle

asmigala:


shawkins:
Is this something that you validated with GSS?  In this case the error message from oracle is erroneous - it is not a recursive common table expression, but a bug when the common table name matches the current schema name.  Also support for the column list depends greatly on the specific oracle version - even when recursive common tables are supposed to be supported with 11.2 - our test instance of 11.2.0.2 does not allow a column list.  However it does look like the column list is definitely required for oracle 12+ so a change similar this with will be needed once all of the version specifics get hashed out.

asmigala:
Reopened as #667 with correct JIRA issue

PR #667, opened 2016-04-13
TEIID-4134 Fix recursive CTE pushdown with Oracle

asmigala:
This might not be sufficient fix, due to possible issues with some specific Oracle 11g versions.


shawkins:
If possible I'd also like to get the update to the pushdown version in this change as well.  If we can't find a better answer, the it should be set to 11.2.0.4 since we have confirmation that will work.

PR #669, opened 2016-04-15
TEIID-4142: update loadmatview as a universal procedure to update Mat View

kylinsoong:
update loadmatview as a universal procedure that can update both Internal and External Mat View

### DIFF
--- a/engine/src/main/resources/org/teiid/metadata/SYSADMIN.sql
+++ b/engine/src/main/resources/org/teiid/metadata/SYSADMIN.sql
@@ -150,6 +150,15 @@ BEGIN
 	END		  
 
 	DECLARE string statusTable = (SELECT "value" from SYS.Properties WHERE UID = VARIABLES.uid AND Name = '{http://www.teiid.org/ext/relational/2012}MATVIEW_STATUS_TABLE');
+
+        /* statusTable is null hints View is Internal Mat View*/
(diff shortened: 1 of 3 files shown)


vhalbert:
Looking at:
      /* statusTable is null hints View is Internal Mat View*/
+        IF (statusTable IS NULL)

An external materialized table isn't required to define a status table, will that cause an issue here? (unless this has changed in 9.x)

shawkins:
> An external materialized table isn't required to define a status table, will that cause an issue here? 

An external materialized view is required to have a status table.

vhalbert:
Did this start in 9.x or is this also required in 8.12.x?   I asked, because reading the doc's, there's nothing indicating "required" as there are for other settings.

shawkins:
No, this is not new - https://docs.jboss.org/author/display/teiid812final/External+Materialization optional=false

kylinsoong:
@vhalbert, if you look the external document [1], In extension properties section, `teiid_rel:MATVIEW_STATUS_TABLE` optional is false. Also as [2] around line 77, if statusTable is null throw TranslatorException

[1] https://teiid.gitbooks.io/documents/content/caching/External_Materialization.html
[2] https://github.com/teiid/teiid/blob/master/engine/src/main/java/org/teiid/query/metadata/MaterializationMetadataRepository.java

PR #680, opened 2016-04-22
Fix unit test

vhalbert:


shawkins:
This has already been committed.

vhalbert:
I got latest and it didn't appear to be there

shawkins:
It's there https://github.com/teiid/teiid/blob/63-8.12.x/connectors/translator-infinispan-dsl/src/test/java/org/teiid/translator/infinispan/dsl/metadata/TestAnnotationMetadataProcessor.java

PR #692, opened 2016-05-19
TEIID-3509: update updateMatView as a universal procedure to refesh mat view by rows

kylinsoong:
Use the similar logic like query status and load view, `SYSADMIN.updateMatView` compatible with internal's `SYSADMIN.refreshMatViewRow` and `SYSADMIN.refreshMatViewRows`, what updateMatView  plan as role as a universal procedure to refesh both internal and external mat view by rows


kylinsoong:
@shawkins could you take a look at this request? thanks for any comments in advance.

shawkins:
Thanks Kylin.  Yes, I'll take a look.

shawkins:
This looks good.  A couple of minor things.  The whole operation should be in a begin atomic block.

Also creating the key with:

pkcolums = pkcolums || ', ' || colname.Name;

Can be an issue when the column name is a reserved word - but it should always work to use table.columnname as we still have the restriction that column names can't contain '.'.

We can build on this later to internally handle a predicate based update or to batch the updates, but for now this is good.

kylinsoong:
Thanks @shawkins, 

> Can be an issue when the column name is a reserved word - but it should always work to use table.columnname as we still have the restriction that column names can't contain '.'.

I am not completed understand, in my understand, the metadata loading will validate, if a column name is a reserved word, metadata loading will throw a exception, so could you give me a example?
Kylin

shawkins:
If a primary column name is for example TABLE - it either needs to be quoted as "TABLE" or qualified as tbl.TABLE for the parser to understand that it's an identifier and not the reserved word.

kylinsoong:
Thanks @shawkins , I got it, and do a minor change.

PR #716, opened 2016-06-09
TEIID-4256: Adding the Date support to Hive, Impala still does not su…

rareddy:
…pport Date


shawkins:
The only change would be to just remove the simpledateformat and use the formatDateValue function.

PR #726, opened 2016-06-20
TEIID-3617: Provide an option to Limit per user based connections

kylinsoong:

PR #741, opened 2016-06-29
TEIID-3617: add an option to limit session counts per user

kylinsoong:


rareddy:
I think I still do not like the way it is designed (that could be just me)
- I do not like single property for managing all the users, I rather define row for each user
- Do not do user management in VDB Service, AuthManager stuff, I would have liked to see more of that in the session service.

kylinsoong:
> I do not like single property for managing all the users, I rather define row for each user
Why? what's the benefit of use the 'define row for each user'? I see use  a single property in many place, even in jboss-admin, createDataSource, you use a single property for all connection-properties.

> Do not do user management in VDB Service, AuthManager stuff, I would have liked to see more of that in the session service.
Not manage user, it just add the VDB lifecycle listener.  Do we need manage the user in session service Scope(system Scope)? Session service is system level, we only have one session service for for transport, in https://github.com/teiid/teiid/pull/726 I do in this approach, but we abandon that way after several discuss and commits, Do we still need roll back that way?

Lastly, do we really need TEIID-3617? we have max-session-limit? do we still need  max-session-limit-user? TEIID-3617 raise by a community user, but it not our commercial customer, we may just ignore it, or let it open to community, do you think it's fine? if yse I will close this pull request, and set TEIID-3617 open to community.

PR #749, opened 2016-06-30
TEIID-3862 adding datasource scripts for excel

vhalbert:


rareddy:
Can also submit for master please?

PR #752, opened 2016-07-04
TEIID-4311: add create XA datasource

kylinsoong:


rareddy:
We do not need to introduce another Admin method, we need to use the same "createDatasource" method. The template name is the differentiation. You duplicated much code from previous method, which is not required.

kylinsoong:
> The template name is the differentiation?

Thanks for your comments. What's the template means in your comments? I also a asked this question in https://issues.jboss.org/browse/TEIID-4311, why create a no-xa/xa datasource in wildfly need a template? In my understanding create a datasource basically need 3 steps:
1. add driver module
2. create jdbc-driver base on driver module
3. create datasource
note that, step 3 we have method in Admin, but 1 and 2 not. I never try to use a template to create a data source.

> we need to use the same "createDatasource" method.

Why use the same method? xa and no-xa are different, there have different properties, no-xa can be create base on one add cli, xa need a composite operaton, which means several add cils in a batch. Also in WildFly, no matter web-console, or management cli, both use different ways to create xa/no-xa datasource, why we try to use same method in teiid.

Thanks
Kylin

rareddy:
Not there yet ;) get rid of "initXADSclassSet" we never hard code the class names. Juts like in "isXADataSource", when there is XA data source class in the driver definition, introduce a new template name ending with "-xa", that is key to determine a data source is XA or not with READ ONLY property of data source name. You would need to "add" this as "read only" property into template properties list. I have mentioned in my comment in JIRA originally.

Get rid of any space only changes, see if can squash commits into single commit.

kylinsoong:
> get rid of "initXADSclassSet" we never hard code the class names. 
DONE

> introduce a new template name ending with "-xa"
Refer to JIRA for more confirm.

>Get rid of any space only changes, see if can squash commits into single commit.
I think it's should be better wait some round, I am afraid I may need do squash commits into single commit again if I do it right now.

PR #753, opened 2016-07-04
TEIID-4297: add OSDQ function

kylinsoong:
>NOTE: This not finished, but any comments/advice will be appreciate.


kylinsoong:
This request will close, @SnailJie will do build script related works.

rareddy:
See you can add them into separate files then  'SourceSystemFunctions' and 'FunctionMethods' and 'SystemSource' to give this library a clean separation from core Teiid stuff. 

Also, also try using the TeiidFunction annotation to define the functions. Where the module dependency for the kit?

Can not accept the way it is, please change accordingly submit again.

PR #760, opened 2016-07-12
TEIID-4277 fixes issues related to returning incomplete results when materialization was occuring

vhalbert:
TEIID-4277 fixes issues related to returning incomplete results when materialization was occuring; synchronized the proxy, handled restarts by not resetting alias cache if already initialized and add connector logic so that it knows which cache to pull during materialization


vhalbert:
will create a separate PR for 9.0 due to the upstream refactoring.

PR #761, opened 2016-07-13
TEIID-4277 fixes issues related to returning incomplete results

vhalbert:
TEIID-4277 fixes issues related to returning incomplete results whenmaterialization was occuring


shawkins:
I am uneasy about this approach.  Having implicit redirection intimately tied to the notion of the materialization cycle seems fragile and is not generally correct (as the user could also be targeting the materialization table for updates outside of the refresh procedure).  Can we just have an explicit staging cache table target?  Or as we talked about we can also promote a refresh approach based upon updating the object model to include a timestamp/version column.

vhalbert:
regarding explicit staging cache table, since JDG doesn't currently support cache renaming, that becomes an issue when its time to swap out the caches.

shawkins:
> The problem is there's no logical transaction grouping the steps (BEFORE, LOAD and AFTER) when the Teiid engine perform materialzation for which the translator knows about in order to control which cache its to use.

That's what I'm trying to avoid.  I don't want the translator to have an intimate knowledge of what the materialization process should be.  It should just expose the necessary operations.

> regarding explicit staging cache table, since JDG doesn't currently support cache renaming, that becomes an issue when its time to swap out the caches.

No that wouldn't be an issue - the naming logic you have acts effectively as a table alias(es).  The rename/swap operation would just switch which table the alias is pointing to.

Couldn't the translator expose Foo table and FooStaging table.  Then on access to either you would use the CacheNameProxy to get the actual cache that each points to.  When swap is called, you simply swap the aliases.  There would then be no special checks for useMaterialization on update, no notion of what state you are at in the materialization process, etc.  But you would have to make truncate operation specific to a cache.

PR #762, opened 2016-07-18
TEIID-4297:Add UDF functions based on OSDQ for POC

SnailJie:
Added systemfunctions: 
String hash(String)
String random(String)
String digit(String)


rareddy:
replaced with pull https://github.com/teiid/teiid/pull/763

rareddy:
Thank you Ja Ren for code

PR #772, opened 2016-07-22
TEIID-4319  deprecating the adding of the class object to the source object

vhalbert:
TEIID-4319  deprecating the adding of the class object to the source table, a holdover from the original design


shawkins:
Can this be done as an import property instead?

vhalbert:
Sure, that make better sense.

PR #776, opened 2016-07-27
TEIID-4319 changed the inclusion of the Object column to be controlled by import property

vhalbert:
TEIID-4319 changed the inclusion of the Object column to be controlled by import property


shawkins:
Merged manually.

PR #777, opened 2016-07-29
TEIID-4307 changed to support the use of NameInSource

vhalbert:
TEIID-4307 changed to support the use of NameInSource for matching to the Pojo, so that the name of the table can be changed to anything the user needs


vhalbert:
documentation coming

PR #782, opened 2016-08-03
TEIID-4333 changed so supportsCompareCriteriaOrdered is set based on JDG version

vhalbert:
TEIID-4333 changed so supportsCompareCriteriaOrdered is set based on JDG version that is determined at runtime based on the version of the cache accessed.  Also, cleaned up libmode to remove old code related to remote cache that is now fully supported under hotrod.


shawkins:
I'm not sure how likely this is, but the version checks may need to be more careful.  Just using a lexical comparison can be a problem - 

"6.6" is greater than "6.10"

If this is possible, you could a version class like in the JDBC translator or make sure that the version string is appropriately padded (such as assume that each component can will be 2 digits - "06.06")

vhalbert:
Can I get this committed then I'll followup with the change related to using similar version class like JDBC?

shawkins:
Sure, let me look things over again.

PR #783, opened 2016-08-03
[TEIID-4183] throw cause of InvocationTargetException

jdurani:
Re-throw cause of InvocationTargetException instead of "swallowing" it so the proper messages can be logged.


shawkins:
It looks like the createSubjectMethod with several arguments is unused in general, can it just be removed?

The if/else that throws Error is not entirely safe, as the exception could just an instance of Throwable.  Could that change just be simplified to logging the cause instead of the InvocationTargetException?

rareddy:
This is class I took from Wildfly, so kept as is.

shawkins:
It looks like this is no longer needed.

rareddy:
sorry, I replaced this with another commit, I thought you were commenting on that pull request

PR #790, opened 2016-08-18
TEIID-4394 - Added analytical function pushdown support for Impala

dkrapohl:


shawkins:
Merged with a minor change to add a version check.  Thanks Don.

PR #791, opened 2016-08-18
Updated infinispan to version 8.0.3

vhalbert:
Updated infinispan to version 8.0.3, which is the version that's in the jdg-7.0 slot in the eap module kits


shawkins:
Shouldn't the property name version.org.infinispan.6 be changed?

vhalbert:
I can update that on this next PR that changes to use Version, like jdbc.

PR #793, opened 2016-08-23
TEIID-4353: feature-pack should as part of the core Teiid build

kylinsoong:
After some investigation and discussion with swarm guy,  feature-pack should be as part of the core Teiid build, feel free if any advice.


rareddy:
As I mentioned before few weeks ago, this looks awfully similar to the Wildfly kit distribution. Unless there are very notable differences other than "resource" vs "artifact" we need to figure out if we can auto-generate these. Maintaining two sets of dependencies is going to be very very cumbersome.

kylinsoong:
Not finished, need adjust feature pack plugin

kylinsoong:
@shawkins Could you look at ths pull request if this can included in Alpha3 release?

rareddy:
@kylinsoong I know you worked hard on this, not trying to be critical, but I do not think we there yet. Few things looking at the commit and not knowing anything about "wildfly-swarm" I observed is

1) Top level module you should have named "swarm" and have the "plugin" and "feature-pack" underneath it. 
2) In "feature-pack" you defined dependencies to Teiid modules fine, but you also defined dependencies to their dependencies looks like which is NOT.
3) I would have imagined, like our "wildfly" kit, every translator, every resource adapter, and teiid modules contribute to the "kit" area with "swarm" specifics then project like "feature-pack" or "build" will pull these into a zip.
4) Most troubled by the "plugin" code, lot of hard coding of paths, modules etc. I feel there got to a better way.

Have you taken a look at other projects as to how they are building their kits, like "keycloak" or "infinispan" for the swarm-feature-pack. Are they doing similar and/or same things as what you are doing in the "plugin" code?

Also, can you explain the relationship between "swarm-teiid" and what is in here? How they are tied and how they work together. If there is document some on swarm site where they explain this please show us that, so that we can better understand this.

kylinsoong:
I am in preparing a sides, let me do some introduction is next meeting.

PR #803, opened 2016-09-26
TEIID-4472 adding cli scripts for installing into ha and full-ha

vhalbert:
TEIID-4472 adding cli scripts for installing into ha and full-ha configurations


shawkins:
Why is the salesforce-34 commented out?

vhalbert:
Not sure, I just took the standalone cli script and duplicated it and then made the caching changes.  It was already commented out.

shawkins:
That is probably from the initial implementation that used the slot mechanism.  It should be fine to have them both defined now.

PR #807, opened 2016-10-06
TEIID-4478 Missed one location where it should have been calling the new findMethod

vhalbert:
Missed one location where it should have been calling the new findMethod(..) so that it handles any case of names


shawkins:
what jira is this under?

vhalbert:
TEIID-4478

PR #808, opened 2016-10-13
TEIID-4511 add st_intersection function

asmigala:


asmigala:
@shawkins the integration tests currently fail due to the added function, it appears the change is not isolated (the tsid changed on many other functions as well), is there a way to regenerate the expected results in bulk, or does that have to be done manually?

shawkins:
You can use -Dreplace_expected=true as an argument to the maven run

That overrides a flag on TestMMDatabaseMetaData that generates most of the results files.

shawkins:
Thanks Andrej.  I expanded this a little to update the pushdown for Teiid/PostgreSQL.

PR #810, opened 2016-10-17
TEIID-4519 prevent execution of updates and deletes on child objects

vhalbert:


shawkins:
Is there a test that can be included for this?

vhalbert:
This was done interactively, so I'll create the tests.

shawkins:
Closing as per the JIRA, since this will not be applied upstream.

PR #811, opened 2016-10-17
TEIID-4510 fix the method getMaxFromGroups()to return 1

vhalbert:
TEIID-4510 fix the method getMaxFromGroups()to return 1, so that each table is independently queried and the results are joined in the teiid engine


shawkins:
Do you mean this as a temporary change?  It looks like there is some logic that is trying to deal with the notion of parent/child relationships for each translator.

vhalbert:
the logic assumes it will deal with only 1 level (table) at a time, so the max group should have been 1.  However, the logic must still distinguish between parent / child because the children will operate differently if its a 1-to-1 versus a 1-to-m (list, array, etc.).

PR #814, opened 2016-10-19
TEIID-4455: Change planning step for from_unixtime() pushdown

kylinsoong:


kylinsoong:
@shawkins , thanks for your time, while run TestFunctionPushdown/testFromUnitTime(), I have found it can go into CapabilitiesUtil.supportsScalarFunction's 
~~~
        if (schema == null || !schema.isPhysical()) {
            // Find capabilities
        	
            if (!caps.supportsFunction(fullName)) {
            	//special handling for delayed rewrite of concat2
            	return (schema == null && SourceSystemFunctions.CONCAT2.equalsIgnoreCase(fullName)
                		&& caps.supportsFunction(SourceSystemFunctions.CONCAT) 
                		&& caps.supportsFunction(SourceSystemFunctions.IFNULL)
                		&& caps.supportsCapability(Capability.QUERY_SEARCHED_CASE));
            }
~~~
and false return,  but I aware this logic only for one function concat2, so my concern is if there are lots of functions need to add to push down, these section will become lots of if/else statement, do you think it's fine we add  from_unixtime logic here?

kylinsoong:
@shawkins , I have do some change, could you review the change?

shawkins:
These changes look good for supporting from_unixtime pushdown, but we need to go further to also support the old behavior of allowing pushdown if timestampadd is supported as well.

Additional changes:

- Add from_unixtime to the hive translator as well.
- Update CapabilitiesUtil.supportsScalarFunction similar to concat2 to check for timestampadd if from_unixtime is not supported
- Update the LanguageBridgeFactory to convert to timestampadd if from_unixtime is not supported

kylinsoong:
@shawkins, could you re-review the changes? thanks.

shawkins:
I think this looks good.  Can you squash to a single commit and also add a test that shows the pushdown of from_unixtime if just timestampadd is supported?

kylinsoong:
https://github.com/teiid/teiid/pull/815 has been created which contains all changes.

PR #815, opened 2016-10-25
TEIID-4455: Change planning step for from_unixtime() pushdown

kylinsoong:


kylinsoong:
This is the pull request migrated from  https://github.com/teiid/teiid/pull/814

PR #816, opened 2016-10-26
TEIID-4515: add Phoenix/Hbase support offset

kylinsoong:


shawkins:
The only potential issue I see is that the Phoenix driver hard codes the server version: https://github.com/apache/phoenix/blob/master/phoenix-core/src/main/java/org/apache/phoenix/jdbc/PhoenixDatabaseMetaData.java#L616

So the automatic version detection will be fooled if a newer driver is used with an older server.

We can either look for a sql query to detect the version, or simply note this possibility in the documentation so the user will know to manually set the database version if there are issues.

kylinsoong:
I am not find a appropriate functions(http://phoenix.apache.org/language/functions.html) to detect the version, probably  document may is a choice?

shawkins:
Ok, we'll note that in the docs.

kylinsoong:
> Ok, we'll note that in the docs. Do you want me to add a note?

Thanks @shawkins , I have created a https://github.com/teiid/teiid-documents/pull/78 about to the document, could you take a look at? feel free to add notes if you want.

PR #818, opened 2016-10-31
TEIID-4545: add phoenix string, time, numberic function as supported/pushdown function

kylinsoong:


shawkins:
Kylin this looks pretty good.  A couple of notes:

PARSEDATE and PARSETIME aren't in SourceSystemFunctions because they are rewritten as cast(parsetimestamp() as date/time).

Changing the HBASE constant to phoenix should not be done in isolation.  We'd probably want to rename the whole executionfactory/translator.  See the SAPNWODataExecutionFactory for an example of that.

Other than those concerns, just make sure that there's no version dependent issues with the function support.

kylinsoong:
Thanks @shawkins, I have spend several hours to review phoenix release notes, mainly reviewed the version 4.2.1, 4.5 and 4.8, seems all functions referred in ef are all supported.

Though hbase ef can mapping the table in hbase(via name in source), the functions, joins, etc all are point to the phoenix, so I have add another PhoenixExecutionFactory.

Any comments/advice? thanks your time again.

shawkins:
You'll need to have HBaseExecutionFactory extend PhoenixExecutionFactory and add the deprecated attribute to the translator annotation on the HBaseExecutionFactory

kylinsoong:
A new request(https://github.com/teiid/teiid/pull/819) be created which only focus on phoenix function pushdown support. Let's move on deprecate HBase ef to separate jira

PR #819, opened 2016-11-03
TEIID-4545: add phoenix functions as supported/pushdown functions

kylinsoong:


shawkins:
This is getting closer.  When adding a translator there are unfortunately a number of updates that are needed.  Including to the standalone-teiid.xml config, the domain cli script, and an entry added in the META-INF/services.

In the translator annotation, you should add:

@Translator(name="hbase", deprecated="phoenix"...

Which adds into the metadata read by the cli/ui that phoenix should be used instead.

Ideally for full backwards compatibility the namespace used for the hbase translator functions should remain hbase, rather than becoming phoenix.

kylinsoong:
Thanks @shawkins, 
> Ideally for full backwards compatibility the namespace used for the hbase translator functions should remain hbase, rather than becoming phoenix.

Before this issue, hbase translator doesn't define any push down functions and supported functions, for the deprecated HBaseExecutionFactory, I think all contents can be removed, which like SAPNWODataExecutionFactory

shawkins:
Absolutely correct Kylin.  Ok, so then we are good with these changes.  Can you rebase to the swarm changes in master and add an entry in the release-notes?

kylinsoong:
I have met some problem in merge the swarm changes, so re-submmit a new request 826, this one will close.

PR #820, opened 2016-11-03
TEIID-2578: adding the authorization support

rareddy:


shawkins:
Merging with the changes:

- added hasAdminRole to CommandContext and made the logic consistent with how we get the datapolicy roles
- Changed the name of Permission.Allowance to Permission.Privilege
- Removed the check of any authenticated role from allowDDLEvent.
- The DUMMY_UPDATE_PLAN should return a single 0 row to be consistent with our other ddl/non-dml statement handling.  However to be fully correct we may need to return -1, but the 0 assumption exists in multiple places.

there is still overlap though between some data role permissions, for example when altering a view definition, that we'll need to resolve.

rareddy:
Sounds good. 

- The DUMMY_UPDATE_PLAN is resulting in assertion error in TupleBuffer. Can you ran Integration Tests and you will see the behavior.

- how is any authenticated supported, or it should only apply to the data?

shawkins:
- my mistake with DUMMY_UPDATE_PLAN was that I had the start index as 0.  It should have been 1, but the value needed to stay 0 (or more correctly there and several other places it should probably be -1)

- any authenticated means that the role is applicable, but there was no check that the role actually granted any relevant permission.  So as of now it's just checking for the admin role for all ddl operations, but that should be refined.

PR #829, opened 2016-11-11
Do not get rid of an excepiton in BasicCryptor.decrypt()

hokuda:


shawkins:
I believe the original intent of not rethrowing the exception was to not reveal the entire stacktrace in case there were version dependent exploits.  As you can see in the source all of the exceptions that a client may trigger on the server side do not rethrow the original exception.  However since this class is used in both the client and the server we could make the client usage rethrow all exceptions.  Is there a reason this change is needed - in general this needs linked to a JIRA issue.

hokuda:
From developer perspective, a verbose stacktrace might be unnecessary. But from support perspective, it is sometimes only a clue to get close to a root cause of a cutomer's issue. I will file a JIRA next week, since it is not functioning now.

shawkins:
Closing this to work toward a resolution on the JIRA

PR #830, opened 2016-11-14
TEIID-4570: fix rest war generate failed issue

kylinsoong:


rareddy:
Replaced yours with this https://github.com/teiid/teiid/pull/831

PR #834, opened 2016-11-18
TEIID-4571: Adding a delegation capable SPNEGO Login Module

rareddy:


shawkins:
It looks like the SPNEGOLoginModule change was also addressed with SECURITY-910.  I'd feel better to push for the login module / delegation changes to be back ported.  Then we can be more consistent in the documentation.

rareddy:
Yes, you are right, Looks like this is already been fixed in https://bugzilla.redhat.com/show_bug.cgi?id=1267738 in 6.4.3 but not in the patch release notes, looking at https://access.redhat.com/support/cases/#/case/01508497 it may be slated for either 6.4.4 or 6.4.5

rareddy:
I will try to pull in changes related verbose addition of  private credential in Teiid code from this pull in master to clean up the code.

PR #841, opened 2016-12-01
TEIID-4441: Add FTP/SFTP based access to the File Resource Adapter

kylinsoong:


rareddy:
Few Questions
1) Why there is a separate FtpsManagedConnectionFactory? why is this not folded into FtpManagedConnectionFactory with additional flags? You know that our tooling can not support, nor ra.xml does not support having two different MCF files in single resource-adapter?

2) Why not update the FileConnection to include the delete, filelist etc, rather than introducing separate FtpFileConnection?

3) What is need for "rename"? May be this useful also in File Translator?

4) If you are going to almost have same metadata for access, then why not have same metadata? i.e remove the difference between getFile vs getFTPFile etc. Unless I am not seeing something. This goes along with (2), have single abstraction of File stuff.

5)  I see you are creating "temp" files when you read from the ftp site, note that you are not obligated to re-use the File stuff, it is easier to return a InputStream to avoid this I think I am fine with that. May be we should look at JBoss VFS system, I know we have similar need in supporting HDFS. That can be the abstraction layer?

Rest looks good.

kylinsoong:
Thanks @rareddy , it's good questions.

As I start to develop ftp ra, the first thing I thought is not change API and it's implements, this can almost answer your Question 2 and Question 3, if add more methods on `FileConnection`, then the `FileConnectionImpl` under connector-file also need some change.

For Question 3, just like add, delete, rename also is a frequent operations for ftp server.

For Question 1, I will merge two mcfs to one.

For Question 5, I would like to investigate more about JBoss VFS now.

rareddy:
Go ahead move comment about the VFS to JIRA, as that may change all the other questions. May be we need something parallel to File translator. I want to get shawkins input on that.

shawkins:
Having the same set of metadata for common operations would be good.  Then we can have procedures that are specific to the other sources as well.  A consideration is that the current getFiles procedure supports only a single directory and only supports the * wildcard.  We may want more or less functionality depending upon the source, but any change to the file translator must be backwards compatible, or introduce a new procedure.

kylinsoong:
The content be move to https://github.com/teiid/teiid/pull/855

PR #842, opened 2016-12-01
OData: Add support for servers not handling comparisons for boolean functions.

NKame:
In OData, some old servers choke on "startswith('a', 'b') eq true"
whereas "startswith('a', 'b')" is fine (same for all the functions returning a boolean). This patch adds a configuration option to handle those cases, if the option is unset the previous behavior is unchanged.

The diff view in GitHub seems broken, what did I do wrong?

### DIFF
--- a/connectors/translator-odata/src/main/java/org/teiid/translator/odata/ODataExecutionFactory.java
+++ b/connectors/translator-odata/src/main/java/org/teiid/translator/odata/ODataExecutionFactory.java
@@ -75,6 +75,7 @@ public class ODataExecutionFactory extends ExecutionFactory<ConnectionFactory, W
 	private boolean supportsOdataCount;
 	private boolean supportsOdataSkip;
 	private boolean supportsOdataTop;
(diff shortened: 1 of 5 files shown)


rareddy:
First you need to open a JIRA @ https://issues.jboss.org/projects/TEIID

Then fork the github project, and submit your patch against "master" branch, then we can take look at the patch. IMO there is no reason to have configuration to enable this option, it should work either or.

NKame:
I think it should be configurable because starting from OData v2, all the examples specify the usage of eq or ne operators when using functions like startswith or endswith, it's really the special case of an server based on a old version of odata4j which does not support that syntax.

rareddy:
ok, fine. I thought this was in OData V4 which allows it.

NKame:
https://issues.jboss.org/browse/TEIID-4613

I'll rebase the patch.

rareddy:
Make sure they show up as additions to existing files, not replacing them

NKame:
Closing the PR to open a new one with a proper patch.

PR #851, opened 2016-12-06
TEIID-4622: Adding new data quality functions from osDQ

kylinsoong:


rareddy:
The methods look like JAVA methods, can we rename to be more like SQL methods. Some suggestions

1) encryptStr = encrypt
2) decryptStr = decrypt
3) encryptArrayStr = encrypt (name overloading)
4) whitespaceIndex = LOCATE function already has same functionality
5) isValidCreditCard = validatecreditcard ?
6) ..

shawkins:
The decrypt/encrypt functions need to be specific in their name as to the algorithm and key length, i.e. encrypt_AES256 - this ensures that we can pushdown when possible without a lot of extra work in the engine/translator to understand what is supported based upon argument values.  These types of functions also typically accept binary values in addition to strings.  We need to understand more about where the is stored, how it is generated, etc.  It's more typical for databases to have functions that pass in the key.  We may come to the conclusion here that we directly want to own these functions - like the recently added hash functions.

whitespaceIndex - isn't quite the same as the locate operation since it's looking for a character class, rather then a specific string.

epochToDate is redundant with from_unixtime

dateToEpoch without a timezone for consistency should be the function to_unixtime.  With a timezone is slightly confusing as we're dealing with the UTC value already, so we'll need to check the source as to what it's doing.

diffInSec is redundant with TIMESTAMPDIFF using the second interval

convertToFormat is likely redundant with formatDate.

The scalar forms of stats functions (median, range, variance, stdDev) may need more discussion - or at least examples showing when they would be called with array values.  They could be aggregate functions (of which we currently have stddev_pop, var_pop, stddev_samp, var_samp).

kylinsoong:
* The decrypt/encrypt function, the osdq used key length is 16, so I change the decrypt/encrypt function to decrypt_AES16/encrypt_AES16. 
* Change some function's name more like SQL methods, rather than java methods
* Remove redundant Time related function

shawkins:
After further review, I don't think we should use the encryption/decryption functions - there are a lot of assumptions that are built-in around the string handling, which does not match any other crypto functions in databases.  We should instead add encrypt/decrypt functions that are consistent / compatible with the functionality provided by pgcrypt, mysql, etc.

We should also leave off the stats functions for now and instead log an issue to add additional aggregates, such as range and median (although the former can be computed with lead/lag which may be added by a separate issue and the latter can use a ranking window function).

kylinsoong:
Thanks @shawkins , I have removed stats, decrypt/encrypt functions

shawkins:
I think this looks good.  In a follow-on issue we should consider adding a new function category and also we may want to think about having these functions reside in a particular schema.

PR #855, opened 2016-12-14
TEIID-4441: Add FTP/SFTP based access to the File Resource Adapter

kylinsoong:
I have 2 concerns of current request:
1. I add a `VirtualFile[] getFiles(String namePattern)` in VirtualFileConnection, it looks redundant?
2. The ftps still not work,.


rareddy:
I like the most of the commit, but I can't seem to get over the extra translator concept "vfsfile". That seems verbose. The current translator "file" should be extended for "VirtualFile". To do this, basically current FileExecutionFactory uses, FileConnection which extends Connection. VFSConnection also extends the Connection. So, make the FileExecutionFactory based on Connection, and then check for the type of connection to modify the behavior of translator. That way we have single translator that is "file"

kylinsoong:
@rareddy 
The issue I referred in last catch up meeting still exist.
Current ftp VirtualFile is a String file, which the stream get by FTPClient, but FTPClient.retrieveFileStream must be call complete command after stream be closed:
~~~
FTPClient client = ...
try {
InputStream in = client.retrieveFileStream("pathToFile");
...
in.close()
} finally {
client.completePendingCommand();
}
~~~
More details from [1], especially the sentence "To finalize the file transfer you must call completePendingCommand and check its return value to verify success"

So if we create a JDBC connection, try to open more stream, there will be a error throw, I think my initial thoughts that cache ftp file to local tmp disk may be good for this issue.

Could you give me some advice.

[1] http://commons.apache.org/proper/commons-net/apidocs/org/apache/commons/net/ftp/FTPClient.html#retrieveFileStream(java.lang.String)

PR #856, opened 2016-12-19
TEIID-4641:Updating the materilization management scripts to use merg…

rareddy:
…e/delete schema with help of new load number column on view


shawkins:
What do you mean in the comments -

update count it not exactly correct, but data is

rareddy:
As mentioned before the update case for the internal materialized case was coded wrong. The criteria based check on matview will find only updated and deleted rows, not inserts. The update count is partial. So, fix this, I had same criteria run against actual view, this will find updates, inserts and will not account for deletes. So that update count is also partial in this case. But we need both to be run to be correct in terms of data. But side effect is updated rows are counted twice.

shawkins:
Ok, I changed that to union the pklists so that we are dealing only with a distinct set for the update.  I think it looks good from there.

PR #857, opened 2016-12-21
TEIID-4643: Add encrypt/decrypt functions

kylinsoong:


shawkins:
Thanks Kylin, this is a good start.

I was thinking that we would want to specify as much as possible in the function name to ensure ease of pushdown support, but there are four factors for encrypt/decrypt - algorithm, block/key size, block mode, and padding.  That's quite a bit to specify in the function name - for example encrypt_AES128_CBC_PKCS5Padding.  So being more general, similar to the function names that you have, may be the easier route and we'll worry about pushdown support later.  

A small thing is that based upon mysql/postgresql functions, the typical convention is data first, then the key.

Proper support of CBC either requires allowing the initialization vector to be passed into the function or the use an explicit initialization vector - see https://issues.jboss.org/projects/TEIID/issues/TEIID-4663 which prepends a random block.  https://www.postgresql.org/docs/9.3/static/pgcrypto.html pgp_sym_encrypt is similar but follows a pgp specific packet format.

There's an extra base64 encoding/decoding being preformed that is unnecessary.  The result of encrypt should be the result bytes, not the bytes of the base64 encoding.

kylinsoong:
@shawkins 
I have added and renamed functions as
~~~
encrypt_AES128_CBC_PKCS5Padding
encrypt_AES128_ECB_PKCS5Padding
decrypt_AES128_CBC_PKCS5Padding
decrypt_AES128_ECB_PKCS5Padding
~~~
I also use common-core api(org.teiid.core.crypto.SymmetricCryptor), do you think it's fine?


org.teiid.core.crypto.SymmetricCryptor

shawkins:
The method names definitely do get too long, let's simplify to something closer to your original methods:

aes_encrypt(data, key)
aes_decrypt(data, key)

We'll document that this uses 128 bit, PKCS5Padding, and CBC with an explicit initialization vector - all of which is done by the SymmetricCryptor.  However a change is needed to explicitly pass in the transformation/algorithm to the SymmetricCryptor rather than to rely on the internal class defaults.

kylinsoong:
Thanks Steven, I have some concerns
* Do we need still consider both CBC and ECB? if yes, we may add another parameter aes_encrypt(data, key, ECB), aes_decrypt(data, key, ECB)
* SymmetricCryptor only contain a protected construct, I do wonder we have to expend it to enable pass the algorithm

shawkins:
> Do we need still consider both CBC and ECB?

Initially no. Since pushdown will be an issue regardless, we should just start by offering CBC.

> SymmetricCryptor only contain a protected construct, I do wonder we have to expend it to enable pass the algorithm

Yes, we'll want to expand the constructor to make things explicit.

PR #879, opened 2017-01-24
TEIID-4718  added feature to enable HotRod client to use SSL

vhalbert:


shawkins:
I don't see the sni host name being used.  Also the password fields should use the $masked extension and may need to be doc'd as using an encrypted value.

This may also be a doc concern, but are there any considerations about the file names - will the classpath, local filesystem, something relative to eap, etc. be used.

vhalbert:
As for SNI, it appears that this feature was added for JDG 7.1 and those jar's are not available at the moment.  So the properties are there, just not used yet.  

As to location, the initial code will take what's given, meaning a full path or relative path (assuming JDG will find it) can be specified.     If you saying that maybe Teiid should offer options for building the String file name that's passed, then sure.  I'll log an enhancement.

As for the $masked extension, I'll change that.

shawkins:
One more small thing, can the direct sys outs be removed from the tests?  A logging flag or just java logging could be used instead.

shawkins:
Thanks Van

PR #884, opened 2017-01-30
TEIID-4718  commit to 8.12.x branch

vhalbert:


vhalbert:
This PR should be put on hold while the PR for TEIID-4709 / TEIID-4736 is being done.  This PR will probably have to be canceled because the logic will need changing based on TEIID-4709

shawkins:
Closing for now.  To be resubmitted.

PR #889, opened 2017-02-01
TEIID-4734: Adding safety check to restart unfinished materialiation …

rareddy:
…jobs during a server restart


shawkins:
Without a good bit of extra work this will be a breaking change, so we'll need that called out in the release notes/docs.  This will also only catch a problem when the node comes back up - but it could have been taken offline indefinitely.  We could use the jgroups view to determine members who have dropped to proactively consider a load as in doubt.

rareddy:
I could not think of way to this without a breaking change, as you say it is little convoluted to make it as non-breaking change. I forgot to add to the compatibility notes.  :( 

I did think about the node being off indefinitely, but left as corner case. I could add event listener to the JGroups, but only if we are OK with breaking change.

PR #893, opened 2017-02-03
TEIID-4709 enables JDG authenticatioin and also fixes TEIID-4736

vhalbert:


vhalbert:
This was tested by restarting JDG several times while the user was logged in.  The connection was cleaned up and recreated without impact to the user (user didn't have to relogin).

vhalbert:
noted also, use the PoolBySubject:


                            <security>
                                <security-domain>teiid-security</security-domain>
                            </security>

shawkins:
If there is no subject with the connection request the attempt will just fail correct?

vhalbert:
Default, no.   Default (current) logic doesn't use the username/password.   If the authentication is configured (i.e., AuthSASLMechanism is set) then it will require either a subject or username/password.  And if that is null, you get the JDG authentication exception.

vhalbert:
the defualt usecase is more for out-of-the-box ability to just play with it.

vhalbert:
the PR for the SSL feature will have to be reworked with these change.

PR #896, opened 2017-02-04
TEIID-4476: Upgrade Cassandra Client library

kylinsoong:
Upgrade datastax cassandra driver from version 2.1.3 to 3.1.3, including some api change:
* get version api has a tiny change
* datatype name enum remove the java class type


rareddy:
1) convert list and set types to array type.
2) Why you removed the searchable on index typed fields?

kylinsoong:
> Why you removed the searchable on index typed fields?
The ColumnMetadata are not related with IndexMetadata in latest 3.x driver libraries.

shawkins:
I'll add a change for the index metadata.

On the type conversion I'm fine with using the array mapping, but we may need to look at the binding logic  as we'd probably need to convert to the expected object class.

shawkins:
Logged https://issues.jboss.org/browse/TEIID-4749 as a follow-on to this issue.

kylinsoong:
I will look into this issue.

PR #897, opened 2017-02-07
TEIID-4732 using the column names in inserts

rareddy:


johnathonlee:
added as commit edea999 after testing locally and making release-notes/commit comment changes.  Thanks Ramesh!

PR #903, opened 2017-02-20
TEIID-4778, TEIID-4775: external materialization scheduling issues

rareddy:


rareddy:
Only submit the changes related to the JDG on this ticket, I will have the rest coming as part of other JIRAs based on the Teiid release(s) coming up shortly.

PR #909, opened 2017-02-21
Changes to add security and fix JDG restarts

vhalbert:
This includes: 

- TEIID-4709  supporting JDG security (username/password and passing credentials)
- TEIID-4736  JDG connector doesn't reconnect property when JDG cache (server) is restarted
- TEIID-4758  Permanent materialzation load failure when JDG source goes down

Does not include: TEIID-4718 support for SSL


vhalbert:
The PR has been applied by another PR.

PR #911, opened 2017-02-22
TEIID-2820: Support Couchbase as a resource

kylinsoong:


rareddy:
A initial commit must have some basic inner workings before it can be accepted. I see that you have very good start on the Resource Adapter, but there is nothing on the translator front other than the shell classes. If you are looking for commit point, you can use your own github repo to host the commit and re-iterate on your development until it reaches a certain threshold where the code can be committed to  upstream.

kylinsoong:
I add it as PR, just for quick view the differ between upstream and my branch, and each commit to my branch also show in the PR.

PR #920, opened 2017-03-06
Feat/teiid 4572

amr-alaa-ismail:
Add handling for null values in pivot results that were neglected by default in Solr.
set "facet.missing=true"
Handling the new null values.


rareddy:
Great work. 1) Can you refer to the JIRA in the commit message. Also update the JIRA about all the enhancements made. 2) and squash all the commits into single commit?

amr-alaa-ismail:
Thank you
All commits are now squashed into one commit with the JIRA issue refered in the commit message

I will update JIRA with the enhancements

shawkins:
Looks good.

It looks like support for PARSE/FORMATTIMESTAMP needs to be restricted - supportsOnlyFormatLiterals should return true and supportsFormatLiteral should check the value.  That will allow unsupported formats to be processed in the engine.

The setTransactionSupport(TransactionSupport.NONE); line should remain.

shawkins:
Looking at this in more detail, there are more issues.  The logic assumes a usage pattern - where, group by, what value types mean.  If the actual query deviates from that pattern, then there are failures.  For example:

- The usage of getTimestampElement assumes that a timestamp value must be from a function, but that is not generally correct

- The case of using parse or format timestamp is not accounted for outside of the expected usage pattern

- An error is generated if the where clause is omitted.  Also the date range is determined from just literal values, and is not checking the predicate - so you could have "x = date or y = date" and still satisfy the logic

If it's possible the changes should account for more general scenarios. 
If it's not possible, then there needs to be more validation logic to prevent potentially misinterpreting the query.

amr-alaa-ismail:
Hi all,

We have tried to restrict our solution to match our pattern. Replying to your comments we have done the following: 

- setTransactionSupport(TransactionSupport.NONE);
- "The usage of getTimestampElement assumes that a timestamp value must be from a function, but that is not generally correct" -> now no literals are enabled except in the where clause
- "The case of using parse or format timestamp is not accounted for outside of the expected usage pattern" -> They can be used outside the pattern but with this format parsetimestamp(formattimestamp(timestamp, 'yyyy-MM-dd'),'yyyy-MM-dd' ) 
- "An error is generated if the where clause is omitted." -> Don't see any errors 
- "the date range is determined from just literal values, and is not checking the predicate - so you could have "x = date or y = date" and still satisfy the logic" -> Please I need more expanation

Please check the "newly added" test cases to see if they did what you requested before

Questions:

During test cases there is one case that i couldn't deal with due to "rewriteGroupBy" function. 

        @Test(expected=TranslatorException.class)
	public void formatOfDateFunctions3() throws Exception {
		...
		getSolrTranslation("Select PARSETIMESTAMP(FORMATTIMESTAMP(purchasets,'yyyy-MM'), 'yyyy-MM'), count(*)  from example where purchasets between '2015-08-01 04:00:00' and '2015-10-31 15:13:32.536' group by FORMATTIMESTAMP('2015-08-01','yyyy-MM') ;"));
	
As you will see in my code we tried to limit the usage of functions to match a specific pattern, but if the parameters are given to the FORMATTIMESTAMP as a literal, '2015-08-01', the "rewriteGroupBy" decides to neglect the "groupby" clause. We tried to by pass the rewriteGroupBy, but this method is found in the engine package which is used by whole translators.

shawkins:
Let's split this up a little bit.  There's an upgrade to the solr library that can be separated and easily committed.

On the implementation of the rest of the feature, yes these changes that you've added do help.  However there's more to consider, such as the questions you asked above:

> "The usage of getTimestampElement assumes that a timestamp value must be from a function, but that is not generally correct" -> now no literals are enabled except in the where clause

That's not quite the case I'm referring to.  If you have "group by timestamp_col, ..." the logic will attempt to cast first timestamp value to a Function.  However if the first timestamp value is a ColumnReference, then that will generate an exception.  There is a catch Exception block that general handles class cast exception and other problems, but it would be better to have the logic more exactly handle this check.

> "An error is generated if the where clause is omitted." -> Don't see any errors

Yes, a translator exception is thrown complaining of missing literal values - but now it's an informational error.

> "the date range is determined from just literal values, and is not checking the predicate - so you could have "x = date or y = date" and still satisfy the logic" -> Please I need more expanation

The logic is just looking for two literal values - it does not detect if the predicates are against the same column as what defines the gap, the comparisons involved, etc.  So you could "x = date_literal or y = date_literal" and it would satisfy the logic because there are two literals - but the predicates don't match the usage pattern.

 > We tried to by pass the rewriteGroupBy, but this method is found in the engine package which is used by whole translators.

You should not need to adjust the engine logic.  This is effectively a group by of a literal value, which the engine will remove.

Let's think a bit more about the design of this feature.  It would clarify the logic a lot to use a different function rather than parse/formattimestamp.  A translator can register a pushdown function that is then available for use by any source using that translator.  You could for example have a function like - "GAP(column, 'MONTH')", rather than "PARSETIMESTAMP(FORMATTIMESTAMP(column, 'yyyy-MM'), 'yyyy-MM')".  This also clarifies that the function will only be used for this specific scenario.  The next simplification would be to override the visit(Select) method to implement logic that can look over the whole query, rather than just relying on the default visitation order.

amr-alaa-ismail:
Hi Shawkins,

I have simplified my code to use GAP method. Waiting for your feedback.

shawkins:
Let's separate this out.  Under another JIRA let's work the solr upgrade.  Updating the solr library and noggit is straightforward. However I don't see that jetty is used by solr directly, what is the rationale for that change?

Also zookeeper is effectively shared between the accumulo and solr translators.  We need to verify that there are no compatibility issues on the accumulo side - or look at updating it's libraries as well.

On the new feature, do not add the GAP function to the system library - it would be a SOLR specific pushdown function to signify this expected usage pattern:

select ... gap(col, gap_literal) ... where ... col <= lower_literal and col >= upper_literal ... group by ... gap(col, gap_literal) ...

Any other usage of gap would be disallowed.

In the ExecutionFactory, you can add the function with:

addPushDownFunction("SOLR", "GAP", TypeFacility.RUNTIME_NAMES.TIMESTAMP, TypeFacility.RUNTIME_NAMES.TIMESTAMP, TypeFacility.RUNTIME_NAMES.STRING);

amr-alaa-ismail:
I just change these versions to be compatible with solrJ version 6.4.1
I revert the jetty version to be "9.2.7.v20150116"
I add the gap method to the ExecutionFactory and remove it form the system library

shawkins:
I separated off the version change as https://issues.jboss.org/browse/TEIID-4878

These changes have now been replaced by a new pull request that has been rebased and updated - https://github.com/teiid/teiid/pull/947

Please review the visitor changes and then we can work through any other issues with the execution changes.

amr-alaa-ismail:
Dear All,

I think I am the only visitor. So i don't need to review other changes. So
what will be the next step.

Thanks

shawkins:
You need to review pull request #947.  I believe the SolrSQLHierarchyVisitor are good, but you should ensure that they meet your expectations.  Then the SolrQueryExecution looks like it may need some additional changes - methods like findTimestampIndex seem to built upon assumptions that may need generalized.

amr-alaa-ismail:
Dear Steven,

Could you please specify exactly what method to be improved to start
working on them today as i do 't see anymore enhacements to be done.

Thanks

amr-alaa-ismail:
Dear Steven, 

I have reviewed the changes on pull request [#947](https://github.com/teiid/teiid/pull/947). So please let me know what will be the next step.

Thanks

PR #925, opened 2017-03-13
changed version to 8.12.1.6_4

vhalbert:


shawkins:
It would be best to not restart the numbering - that is start at 8.12.10.6_4

PR #926, opened 2017-03-14
64 8.12.x

vhalbert:


vhalbert:
Would be helpful if the merge/squash buttons were enabled.

PR #928, opened 2017-03-23
fix the NPE seen when the connection is being cleaned up

vhalbert:


shawkins:
Is this still needed for 6.3/6.4?

vhalbert:
6.4.

shawkins:
Added the commits to 6.4

PR #929, opened 2017-03-23
Attempt at improving a bit the join support for the OData v2 translator.

NKame:
It's not possible to completely implement it, but this makes it work for simple cases. It must be activated by supporting innerJoin on the ExecutionFactory, even if the boolean is not explicitely checked in the SQLVisitor (maybe it should).

See discussion [Optimizer and over constraining ](https://developer.jboss.org/thread/274330)

### DIFF
--- a/connectors/odata/translator-odata/src/main/java/org/teiid/translator/odata/ODataMetadataProcessor.java
+++ b/connectors/odata/translator-odata/src/main/java/org/teiid/translator/odata/ODataMetadataProcessor.java
@@ -314,8 +314,17 @@ public class ODataMetadataProcessor implements MetadataProcessor<WSConnection> {
 			else {
 				fromKeys = refConstaint.getDependentReferences();
 				toKeys = refConstaint.getPrincipalReferences();
-			}
+			}			
+			
 			if (matchesWithPkOrUnique(fromKeys, fromTable)) {
+				for(String fromKey : fromKeys) {
+					Column column = fromTable.getColumnByName(fromKey);
+					column.setProperty(JOIN_COLUMN, String.valueOf(true));
+				}
+				for(String toKey : toKeys) {
+					Column column = toTable.getColumnByName(toKey);
+					column.setProperty(JOIN_COLUMN, String.valueOf(true));
+				}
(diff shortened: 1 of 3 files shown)


rareddy:
This needs to go into your own extension, I do not think any where it mentioned/agreed that the Teiid's implementation is at wrong here.

NKame:
It's not really at wrong, but for instance, if you try to do 
"SELECT o.OrderId FROM Shippers s INNER JOIN Orders o ON o.ShipVia=s.ShipperID WHERE s.ShipperID = 12" in Teiid, it's going to generate two OData queries with the standard configuration, and it will fail to generate a valid query if you try setSupportsInnerJoin. Issuing two queries is suboptimal but it works in the general case. But in our case, the special server we must access refuses the two queries approach, because they have some hardcoded patterns and never consider the second member accessed alone. But I perfectly understand that the patch is too specific, I only wanted to avoid maintaining a fork, but that's ok :)

One question that I have is about TEIID17006, I don't understand why a join column must not be selectable?

rareddy:
If Teiid did create a two tables with one-2-one or One-2-Many relation, then two entities are defined with navigation properties, or one entity contains the other. Only in this situations a JOIN clause is allowed/supported such that Teiid can issue a single query. If the Teiid did not generate a JOIN then there is no navigation defined in your source between these two sources, thus two queries are issued. If you want you can add additional metadata to the generated metadata to define the relation in your model and see if that gets you the right query into the translator layer.

Another scenario comes from containment, entity A contains entity B, here the relation is implicit, where Teiid creates FK relation automatically based on JOIN column which is artificial based entity A's identity, thus it is not selectable.

If it bug, I have no problem in accepting it, if this is due to the specific usecase you are dealing with, then I can't. That means, this will be bug for the next user

shawkins:
If this is still needed, it would be best to address it at the engine level:  https://issues.jboss.org/browse/TEIID-4943

NKame:
Thanks for putting the right words on it. Right now we're running on our patched version, since "it works for us", we haven't improved the patch (well, we fixed a small bug that I didn't put in the PR since it's not mature enough to merged).

PR #938, opened 2017-04-06
TEIID-2820: Support Couchbase as a resource

kylinsoong:
This is the replacement of https://github.com/teiid/teiid/pull/921, I have test both via Swarm teiid and Client Server quickstart.


connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseExecutionFactory.java
```diff
@@ -0,0 +1,310 @@
...
+    }
+
+    @Override
+    public boolean supportsAggregatesAvg() {
+        return true;
+    }
+
+    @Override
+    public boolean supportsAggregatesMin() {
+        return false;
```
rareddy: Is this supposed to be true?

connectors/couchbase/connector-couchbase/src/main/java/org/teiid/resource/adapter/couchbase/CouchbaseManagedConnectionFactory.java
```diff
@@ -0,0 +1,309 @@
...
+                return new CouchbaseConnectionImpl(environment, connectionString, keyspace, password, timeoutUnit, namespace);
+            }};
+		
+	}
+
+	@Override
+	public int hashCode() {
+		final int prime = 31;
+		int result = 1;
+		result = prime * result + this.getManagementTimeout().hashCode();
```
rareddy: I do not think you want to include any timeout values to be part of object equality

connectors/couchbase/feature-pack-couchbase/pom.xml
```diff
@@ -0,0 +1,116 @@
...
+            <exclusions>
+                <exclusion>
+                    <groupId>*</groupId>
+                    <artifactId>*</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.teiid.connectors</groupId>
+            <artifactId>feature-pack-jdbc</artifactId>
```
rareddy: Why do you need JDBC translator?

connectors/couchbase/feature-pack-couchbase/src/main/resources/modules/system/layers/dv/org/jboss/teiid/translator/couchbase/api/main/module.xml
```diff
@@ -0,0 +1,19 @@
...
+    <properties>
+        <property name="jboss.api" value="private"/>
+    </properties>
+    
+    <resources>
+        <artifact name="${org.jboss.teiid.connectors:couchbase-api}" />
+        <artifact name="${com.couchbase.client:java-client}" />
+        <artifact name="${com.couchbase.client:core-io}" />
+        <artifact name="${io.reactivex:rxjava}" />
+    </resources>
```
rareddy: create separate modules directories for "com.couchbase" and "io.reactivex"

connectors/couchbase/translator-couchbase/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+            <groupId>org.jboss.teiid</groupId>
+            <artifactId>teiid-api</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.teiid</groupId>
+            <artifactId>teiid-common-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.teiid</groupId>
+            <artifactId>teiid-engine</artifactId>
```
rareddy: Do you need this?

connectors/couchbase/translator-couchbase/pom.xml
```diff
@@ -0,0 +1,66 @@
...
+            <groupId>org.jboss.teiid.connectors</groupId>
+            <artifactId>couchbase-api</artifactId>
+        </dependency>
+        <dependency>
+          <groupId>com.couchbase.client</groupId>
+          <artifactId>java-client</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.jboss.teiid.connectors</groupId>
+            <artifactId>translator-jdbc</artifactId>
```
rareddy: Do you need this? I am trying to keep the dependencies tree to a minimum.

connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseMetadataProcessor.java
```diff
@@ -0,0 +1,718 @@
...
+            return TypeFacility.RUNTIME_NAMES.LONG;
+        } else if (value instanceof Double) {
+            return TypeFacility.RUNTIME_NAMES.DOUBLE;
+        } else if (value instanceof Boolean) {
+            return TypeFacility.RUNTIME_NAMES.BOOLEAN;
+        } else if (value instanceof BigInteger) {
+            return TypeFacility.RUNTIME_NAMES.BIG_INTEGER;
+        } else if (value instanceof BigDecimal) {
+            return TypeFacility.RUNTIME_NAMES.BIG_DECIMAL;
+        }
```
rareddy: How are the Date/Timestamp handled?

connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseMetadataProcessor.java
```diff
@@ -0,0 +1,718 @@
...
+import org.teiid.translator.TranslatorProperty.PropertyType;
+
+import com.couchbase.client.java.document.json.JsonArray;
+import com.couchbase.client.java.document.json.JsonObject;
+import com.couchbase.client.java.document.json.JsonValue;
+import com.couchbase.client.java.query.N1qlQueryRow;
+
+public class CouchbaseMetadataProcessor implements MetadataProcessor<CouchbaseConnection> {
+    
+    public static final String IS_ARRAY_TABLE = MetadataFactory.COUCHBASE_URI + "ISARRAYTABLE"; //$NON-NLS-1$
```
rareddy: Add ExtensionMetadata annotations here for these two properties.

connectors/couchbase/translator-couchbase/src/test/java/org/teiid/translator/couchbase/TestN1QLVisitor.java
```diff
@@ -0,0 +1,492 @@
...
+        sql = "call getTextMetadataDocument('test')";
+        helpTest(sql, N1QL1307);
+        
+        sql = "call getMetadataDocument('test')";
+        helpTest(sql, N1QL1308);
+    }
+    
+    public static enum N1QL {
+        N1QL0101,
+        N1QL0102,
```
rareddy: This seems little excessive, what requires this de-reference and complexity. Could you not simply inline the expected results? Did you need to replace any variables?

connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseMetadataProcessor.java
```diff
@@ -0,0 +1,718 @@
...
+            }
+        } 
+
+        return this.typeNameMap.get(keyspace);
+    }
+    
+    protected void addProcedures(MetadataFactory metadataFactory, CouchbaseConnection connection) {
+        
+        Procedure getTextDocuments = metadataFactory.addProcedure(GETTEXTDOCUMENTS);
+        getTextDocuments.setAnnotation(CouchbasePlugin.Util.getString("getTextDocuments.Annotation")); //$NON-NLS-1$
```
rareddy: I am not convinced that we need these at all, what is the purpose?  Even at that I would like to trim these down to getDocuments and getDocument for direct query process. If the result comes back json text then it clob otherwise blob. These direct access is dangerous. If really want to provide, then look into direct query then these. These should be reserved for sources with no query interface.

connectors/couchbase/translator-couchbase/src/test/java/org/teiid/translator/couchbase/TestN1QLVisitor.java
kylinsoong: > Could you not simply inline the expected results?
This is just for test usage, for make the unit test a little tidy, cause, the N1QL statement may very large, if you look at the N1QL.properties, the length of some expected result are 1000, if we put it inline, the test method looks awful.

> Did you need to replace any variables?
If REPLACE_EXPECTED set to true, all expected results in N1QL.properties will be replace, which same like test integration.

connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseMetadataProcessor.java
kylinsoong: The N1qlQueryResult only contains a Iterable `com.couchbase.client.java.document.json.JsonValue`, which support 12 types are
~~~
public static boolean checkType(Object item) {
        return item == null
            || item instanceof String
            || item instanceof Integer
            || item instanceof Long
            || item instanceof Double
            || item instanceof Boolean
            || item instanceof BigInteger
            || item instanceof BigDecimal
            || item instanceof JsonObject
            || item instanceof JsonArray;
    }
~~~

connectors/couchbase/feature-pack-couchbase/pom.xml
kylinsoong: Yes, I re-used jdbs translator's `org.teiid.translator.jdbc.FunctionModifier` and `org.teiid.translator.jdbc.AliasModifier`

connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseExecutionFactory.java
kylinsoong: I forget to set to return false.

connectors/couchbase/feature-pack-couchbase/src/main/resources/modules/system/layers/dv/org/jboss/teiid/translator/couchbase/api/main/module.xml
kylinsoong: Yes, it make sense.

connectors/couchbase/translator-couchbase/pom.xml
kylinsoong: This should be remove.

connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseMetadataProcessor.java
kylinsoong: This is for providing a write back functionality.
Base on the current schema mapping design, it's possible to add support write back via translate Teiid SQL to N1QL. 

If remove these procedures means big change, I would like you to double confirm, for add a UpdateVisitor, or Direct query, I plan to do in separate PR.

kylinsoong:
## TODO
* Remove teiid-engine dependency
* Add ExtensionMetadata annotations for option properties
* Create separate modules directories for "com.couchbase" and "io.reactivex"
* TBD(remove procedures)

connectors/couchbase/connector-couchbase/src/main/java/org/teiid/resource/adapter/couchbase/CouchbaseManagedConnectionFactory.java
kylinsoong: The timeout is the properties of managed connection factory.

rareddy: I know they are, my comment they are not important for equality or hashcode check.

connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseMetadataProcessor.java
rareddy: so N1Q1 does not support any update features?

connectors/couchbase/connector-couchbase/src/main/java/org/teiid/resource/adapter/couchbase/CouchbaseManagedConnectionFactory.java
kylinsoong: Thanks for point out, AFAIK, two mcf equal means all attribute of two mcf equal, is there any specific to define how to define two mcf equality or hashcode ?

connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseMetadataProcessor.java
kylinsoong: NO, it support
- UPDATE - update an existing one
- UPSERT - insert a new record or update an existing one

https://developer.couchbase.com/documentation/server/4.6/n1ql/n1ql-language-reference/upsert.html

rareddy: then you need to get rid of these and support the directly from transforming SQL statements.

kylinsoong:
@rareddy , Could you have a look? I have tested the PR both with embedded examples and swarm server/client mode, it works fine.
* The GETDOCUMENTS and GETDOCUMENT procedures not removed, the other procedures be removed, these use together with native procedure to execute source query.

rareddy:
I do not have any further comments, looks good. BTW, what is an ArrayTable? How is this different?

kylinsoong:
> BTW, what is an ArrayTable? How is this different?
If a table has an ISARRAYTABLE option
~~~
OPTIONS (NAMEINSOURCE '`test`.`Items`[]', UPDATABLE TRUE, "teiid_couchbase:ISARRAYTABLE" 'true');
~~~
that hints the table is for map JsonArray. Further more, excepts the common type attribute in a json document, there may have other 2 kinds of Object type, nested Json and nested array, the current translator map any nested array to separate table, the couchbase supply lots of array related function [1]
* ARRAY_POSITION - use by current translator to get item index in array
* ARRAY_CONCAT - use by current translator to insert/delete specific item in array
[1] https://developer.couchbase.com/documentation/server/current/n1ql/n1ql-language-reference/arrayfun.html

shawkins:
This look very good. 

A concern in that the N1QLUpdateVisitor.literalValueMapping logic only applicable to insert values.

Update change sets can be defined using expressions:

update tbl set x = y/2 ...

And where clause conditions could be similarly complex:

where x = y/2

If you expect only val = literal comparisons, you will want ExecutionFactory.supportsOnlyLiteralComparison to be true, but even with that change the translator has function support that will also not work with literalValueMapping, such as

round(x, 2) = 10

If you want, I can look at making these changes.

kylinsoong:
@shawkins feel free to do any change.

PR #941, opened 2017-04-20
TEIID-2820 refinements to couchbase updates

shawkins:
I've updated the code to address my initial concerns so that update query construction is more explicit.  An assumption that I've made is that it's valid to use USE KEYS with a WHERE clause - otherwise we are allowing updates to potentially the an unintended document.  I've left some failing test cases to highlight additional scenarios that currently fail.  I'm not sure if those should be expected failures or if the logic should be altered to support those cases.

A minor issue of escaping the quote character in a string literal was also addressed as best I as could guess from the docs.

### DIFF
--- a/connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseExecutionFactory.java
+++ b/connectors/couchbase/translator-couchbase/src/main/java/org/teiid/translator/couchbase/CouchbaseExecutionFactory.java
@@ -42,16 +42,7 @@ import org.teiid.language.Function;
 import org.teiid.language.QueryExpression;
 import org.teiid.language.SQLConstants.Tokens;
 import org.teiid.metadata.RuntimeMetadata;
-import org.teiid.translator.ExecutionContext;
-import org.teiid.translator.ExecutionFactory;
-import org.teiid.translator.MetadataProcessor;
-import org.teiid.translator.ProcedureExecution;
-import org.teiid.translator.ResultSetExecution;
-import org.teiid.translator.SourceSystemFunctions;
-import org.teiid.translator.Translator;
-import org.teiid.translator.TranslatorException;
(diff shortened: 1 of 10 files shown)


kylinsoong:
> An assumption that I've made is that it's valid to use USE KEYS with a WHERE clause 

Yes, it's fine to both use USE KEYS and WHERE together.

shawkins:
> but it meaningless

That is not correct - the additional predicates still need to be accounted for.  If you have pk = val AND other = val1 - simply treating this as pk = val means that you may be allowing modification to the wrong document.

kylinsoong:
> If you have pk = val AND other = val1 - - simply treating this as pk = val means that you may be allowing modification to the wrong document.

The issue is the columns come from nested object in array not be supported well.

* For insert/update/delete array table, the documentID and all index are necessary, a validation logic in N1QLUpdateVisitor
* For query array table, the documentID are necessary, and index are optional,

@shawkins , If you merge this PR, and I can try to add some logic to make sure the junit test run fine.

kylinsoong:
@shawkins , I have done some further investigation and changes in [1], I think your change in this PR/UpdateVisitor have missed some things, a row in a no array table be removed, it's referenced array table item also should be removed.

@shawkins , If you merge or close this PR, I can do more change to make junit test works.

[1] https://github.com/jbosschina/teiid/commit/7c7f5144e5c1f032ce123902ecd36b3e01e05e28

kylinsoong:
@shawkins, I plan to create a new PR which contains the change in this one and fixing the failure junit test for array table as https://github.com/teiid/teiid/compare/master...kylinsoong:teiid-2820?expand=1

Hopefully, I can finish soon.

kylinsoong:
@shawkins , could you close this PR? excepts the teiid-releasenotes.html, all others be included in https://github.com/teiid/teiid/pull/942

PR #944, opened 2017-04-27
TEIID-4733: Adding a Infinispan translator and resource-adapter that …

rareddy:
…understands protobuf & hotrod based schema without explicit java based marshall classes


rareddy:
@shawkins Can you review and pull this in.

connectors/infinispan/translator-infinispan-hotrod/src/main/java/org/teiid/translator/infinispan/hotrod/InfinispanResponse.java
```diff
@@ -0,0 +1,121 @@
...
+    }
+
+    private List<Object> buildRow(Map<String, Object> row) {
+        ArrayList<Object> result = new ArrayList<>();
+        for (String attr : this.projected) {
+            result.add(row.get(attr));
+        }
+        return result;
+    }
+}
```
duchuyctlk: Please add an empty line at end of each source file.

shawkins:
Thanks Ramesh.  Yes, I'll get it in by Monday.

shawkins:
Here's a couple of quick questions/comments:

TestHotrodExecution and TestHotrodServer are ignored - can we document what's missing that would allow them to run?

The unsigned types in ProtobufDataManager will need to map to larger types - or we need an option like jdbc to control how to widen unsigned types.  There are also more types that may need to be mapped, such as SFIXED32

IckleConvertionVisitor should be IckleConversionVisitor

TestProtobufMetadataProcessor.testMetadataProcessorAddressbook does not check the generated ddl

rareddy:
I mapped data types from https://developers.google.com/protocol-buffers/docs/proto#scalar which does not mention the widening, gave the target java types. Do you think we still need it?

rareddy:
I am still having some configuration missing from my test HotRodServer it does not work in embedded mode, thus the commented tests. I enabled "TestHotrodServer" but the "TestHotrodExecution" is commented. There is some injection based stuff missing, could not figure out what :(

Added SFIXED32/64.
Fixed the Ickle classname spelling
TestProtobufMetadataProcessor.testMetadataProcessorAddressbook is just simple test to test for debugging, not testing anything. There are other tests doing validation of ddl.

PR #949, opened 2017-05-01
TEIID-2820 backport of couchbase

vhalbert:
TEIID-2820 backport of couchbase, but will only build if unit test are not run -Dmaven.test.skip=true, those will be fixed in the next commit


shawkins:
The unit test issues were addressed with the merge commit.

PR #950, opened 2017-05-01
TEIID-4495 fixing the back port of changes

vhalbert:


shawkins:
As far as I can tell the only change that is still needed would be the domain cli.

PR #951, opened 2017-05-02
TEIID-4883 Excel Translator Filter Empty Column Cells

dvilaverde:
Enhance the excel translator to filter column cells that are
empty.  This is only enabled when using a header row and when
the importer.allowEmptyHeaderRow is true.

See https://issues.jboss.org/browse/TEIID-4883


connectors/file/translator-excel/src/main/java/org/teiid/translator/excel/ExcelMetadataProcessor.java
```diff
@@ -57,12 +57,16 @@
     @ExtensionMetadataProperty(applicable=Column.class, datatype=Integer.class, display="Cell Number", description="Cell number, where the column information is defined. If column name is ROW_ID, define it as -1", required=true)    
 	public static final String CELL_NUMBER = MetadataFactory.EXCEL_URI+"CELL_NUMBER"; //$NON-NLS-1$
 
+    @ExtensionMetadataProperty(applicable=Column.class, datatype=Boolean.class, display="Allow Empty Column Name", description="Allow the column information to contain empty cells.")
```
rareddy: This is not required, this is only needed if you are adding additional metadata proeprties

connectors/file/translator-excel/src/main/java/org/teiid/translator/excel/ExcelMetadataProcessor.java
```diff
@@ -156,6 +160,22 @@ private void addTable(MetadataFactory mf, Sheet sheet, String xlsName, String or
 		Row dataRow = null;
 		int lastCellNumber = headerRow.getLastCellNum();
 		
+		// only accept cells that have a value for the name when using a header row
```
rareddy: So this is effectively truncating the table definition until the first non empty row. Is that the intention?

connectors/file/translator-excel/src/main/java/org/teiid/translator/excel/ExcelMetadataProcessor.java
dvilaverde: ok will remove

connectors/file/translator-excel/src/main/java/org/teiid/translator/excel/ExcelMetadataProcessor.java
dvilaverde: correct, will only use columns where the header row has non-empty string values.  The first empty string cell in the header row marks the end of the table.

rareddy:
btw, also add tests please.

rareddy:
Thank you @dvilaverde

PR #960, opened 2017-05-24
TEIID-4926: fix the convert defects

kylinsoong:


shawkins:
I'm going to tweak this to not allow the pushdown of date/time/timestamp conversion - as those are not directly supported types.

PR #965, opened 2017-06-02
backporting changes to support better integration with JDG 7.1.

vhalbert:


shawkins:
Is this all you need in this pull request?

vhalbert:
In the teiid project, yes.  The jdg project changes to come, will need these.

PR #969, opened 2017-06-09
TEIID-4781: upgrade mongo-java-driver to 3.x

kylinsoong:


rareddy:
Can comment on why the code changes were required for driver upgrade. Were there any api changes?

shawkins:
Yes, there appear to be api changes.  The only issue that jumps out at me is the auto generated keys logic - there is now only a single upserted id value being returned, but the old logic expected to read individual fields.  What is expected here - only a single ObjectId or are complex keys allowed?

rareddy:
The previous worked with complex keys looks like, but the new one will not.

PR #971, opened 2017-06-19
TEIID-4963: JDBC LocalProfile should not bind with jboss-modules

kylinsoong:


rareddy:
This may not be required, see the setConnectionProfile(ConnectionProfile ..) method, this is supposed to be used in that case. But I don not see why we shouldn't have this either. So, it may help in some situations.

PR #1002, opened 2017-09-05
TEIID-5032 - removing configuration of infinispan translator and resource-adapter

vhalbert:
TEIID-5032 - removing configuration of infinispan translator and resource-adapter will need to run CLI script to integrate JDG


vhalbert:
Can I get this committed?

PR #1005, opened 2017-09-15
TEIID-5065 changing jdbc url regex to accept wider range of characters

pranav-k:


shawkins:
Thanks for the pull request/issue.  I took a slightly different approach so that we could accommodate url encoding as well.

PR #1018, opened 2017-09-25
Backport

vhalbert:
TEIID-4666: Google translator throws SpreadsheetOperationException for query DELETE FROM table (allowing delete/update to work with no predicate).  Also addreses regressions on TEIID-4656, TEIID-4543, TEIID-4667


shawkins:
This seems to represent an intermediate state that we don't want.  It looks like 6.4 matches master including changes from for example https://issues.jboss.org/browse/TEIID-4667 (which doesn't have it's targeted versions correctly set).

PR #1025, opened 2017-10-09
Fix when add @Cache annotation, was missing enclosing paren

vhalbert:


shawkins:
Is it possible to add a test for this?

vhalbert:
Yes, i'll create one.

shawkins:
Is there a JIRA to associate this with?

vhalbert:
This is part of TEIID-5088 and TEIID-5089

PR #1034, opened 2017-11-07
max-active-plans property limit one higher than set

vhalbert:


shawkins:
There is a small risk with backporting this in that I could not determine the original motivation for the change in 7.4.1.

vhalbert:
I will re-verify with QE to make sure.

vhalbert:
QE said to commit it, they will test and confirm.

PR #1041, opened 2017-11-21
TEIID-5165 fixed InfinispanResponse

vhalbert:


shawkins:
Is it possible to have a test for this as well?

PR #1054, opened 2018-01-19
TEIID-5221: Correcting the metadata generation for OData V4 sources

shawkins:
Here's a combined pull request.  There is a failing unit test and just the sys.out from testNorthwind.  Have a look and see what you think of these changes.  It looks like there some fks that were not being created that should be, but creating both the implicit and the explicit can result in duplicate constraints:

CREATE FOREIGN TABLE Orders (
	OrderID integer NOT NULL OPTIONS (NATIVE_TYPE 'Edm.Int32'),
	CustomerID string(5),
	...
	CONSTRAINT Customers_Orders FOREIGN KEY(CustomerID) REFERENCES northwind.Customers (CustomerID),
	CONSTRAINT Employees_Orders FOREIGN KEY(EmployeeID) REFERENCES northwind.Employees (EmployeeID),
	CONSTRAINT Orders_Customer FOREIGN KEY(CustomerID) REFERENCES northwind.Customers (CustomerID),
	CONSTRAINT Orders_Employee FOREIGN KEY(EmployeeID) REFERENCES northwind.Employees (EmployeeID),
	CONSTRAINT Orders_Shipper FOREIGN KEY(ShipVia) REFERENCES northwind.Shippers (ShipperID)
) OPTIONS (UPDATABLE TRUE, "teiid_odata:NameInSchema" 'NorthwindModel.Order', "teiid_odata:Type" 'ENTITY_COLLECTION');

Note both Customers_Orders and Orders_Customer.  That can be avoided by looking for the reverse navigation, but I wanted to run this by you first before going further.

### DIFF
--- a/connectors/odata/translator-odata4/src/main/java/org/teiid/translator/odata4/ODataFilterVisitor.java
+++ b/connectors/odata/translator-odata4/src/main/java/org/teiid/translator/odata4/ODataFilterVisitor.java
@@ -181,16 +181,13 @@ public class ODataFilterVisitor extends HierarchyVisitor {
     @Override
     public void visit(ColumnReference obj) {
         Column column = obj.getMetadataObject();
-        // check if the column on psedo column, then move it to the parent.
-        String pseudo = ODataMetadataProcessor.getPseudo(column);
-        
+
         ODataDocumentNode schemaElement = this.query.getSchemaElement((Table)column.getParent());
-        if (pseudo != null) {
+        // check if the column on pseudo column, then move it to the parent.
+        if (ODataMetadataProcessor.isPseudo(column)) {
             try {
-                Table columnParent = (Table)column.getParent();
-                Table pseudoColumnParent = this.metadata.getTable(
-                        ODataMetadataProcessor.getMerge(columnParent));
-                schemaElement = this.query.getSchemaElement(pseudoColumnParent);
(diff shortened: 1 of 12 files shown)


rareddy:
I see the changes in the implicit case, I did not quite understand how the additional contraints are coming in. Were they created before or with changes you made?

Good call on containment and schema name. Yes, I think we should avoid if we can creating the duplicate navigations.

shawkins:
> I see the changes in the implicit case, I did not quite understand how the additional contraints are coming in. Were they created before or with changes you made?

After.  I split the changes into a separate commit for clarity.  It appeared that in the explicit referential integrity case that the "keyMatches(referenceColumnNames, fromTable.getPrimaryKey())" check was inappropriately filtering because the key was being created in the wrong direction.  Although we can add back such a check if there are cases where the metadata is inaccurate.  I hadn't researched enough, but it seems like we may need to differentiate the key direction by whether the property is a collection.  In the test cases we had though the explicit case was only not a collection.

rareddy:
Great, that was clear now. Changing the key direction is something I struggled with, I saw bunch of test failures so I kind of backed out.

I believe implicit collection cases were the reason I was doing in the reverse direction so that I can treat one2one and one2many similarly. Looks good.

PR #1057, opened 2018-01-25
TEIIDTOOLS-300  change parent dependencies to use new project-bom setup

vhalbert:


shawkins:
Van, should the project pom still have the wildfly and wildfly.core versions?

vhalbert:
Its not perfect, it will take more iterations on the alignment as other projects are updated.

shawkins:
https://github.com/teiid/project-bom/pull/6 should bring us up-to-date with the project.  

It seems like we'll have to release the bom on every master change of a dependency and for each new Teiid version correct?  - That seems like it would include every fix release as well.

shawkins:
Comparing to the wildfly bom and pom, they manage both dependencies and versions in each.  For the ease of community releases I would be in favor of that for now - which would remove the pom changes from this pull request.  Or I would be fine with putting the bom project in Teiid proper and reference it by relative-path so that it can be built/released with Teiid in lock step with the versions.

vhalbert:
Both don't contain the same dependencies (e.g., ironjacammar not in bom).      The bom is a subset of the pom, where the pom appears to contain the global list of versions.

shawkins:
> The bom is a subset of the pom, where the pom appears to contain the global list of versions.

I'm not saying they have to be the same, just that we would need the old dependency - especially the intra-project dependencies back in the project pom.  So I'm advocated that either we remove the pom changes from this pull request, or that we move the bom project into Teiid proper so that it can have its versions managed with the project releases.  Are you ok with either of those approaches, or do you have another suggestion?

shawkins:
Merged with changes to keep the internal dependencies declared in the project pom so that project fix releases are easier to produce.

pom.xml
```diff
@@ -16,104 +16,17 @@
 	<description>Federated SQL and XML query engine.</description>
 	<properties>		
 	    <site.url>http://www.jboss.org/teiid</site.url>
-	    <jdk.min.version>1.8</jdk.min.version>
-	    <maven.compiler.source>1.8</maven.compiler.source>
-	    <maven.compiler.target>1.8</maven.compiler.target>
+	    <!--jdk.min.version>1.8</jdk.min.version-->
```
jstastny-cz: Why has this snippet been commented out? Preferably if this is not needed anymore, it should have been removed.

vhalbert: create a jira.

jstastny-cz: There already is a JIRA about writing down best practices. We need to make sure, this is one of them.

PR #1060, opened 2018-02-01
TEIID-5206: checking precision and scale on EdmDecimal after denormal…

rareddy:
…ization of exponential form


rareddy:
This is for @johnathonlee to pull into 6.4 builds when he is ready to build 6.4.1 or CR6

PR #1065, opened 2018-02-20
TEIID-5206: checking precision and scale on EdmDecimal after denormal…

rareddy:
…ization of exponential form


shawkins:
Already merged.

PR #1068, opened 2018-02-23
TEIID-5253 addressing timing issue when listening for vdb events

shawkins:
# Conflicts:
#	jboss-integration/src/main/java/org/teiid/jboss/MaterializationManagementService.java
#	jboss-integration/src/main/java/org/teiid/jboss/ResteasyEnabler.java
#	jboss-integration/src/main/java/org/teiid/jboss/TeiidAdd.java
#	runtime/src/main/java/org/teiid/runtime/MaterializationManager.java


shawkins:
Yes, looks like this can be closed.

PR #1069, opened 2018-03-02
Fix for build

vhalbert:


shawkins:
Unfortunately that does not work.  It is copying the jar as the war.  Can you account for any other differences in the hudson environment?

vhalbert:
oops.   

I did try changing the maven version, up'ing it to 3.3.9.  but no luck.

shawkins:
Locally my build was using the 1.1.7 of the wildfly-server-provisioning-maven-plugin.  After removing that from the repository it was brought back as 1.1.6.  That matches <version.org.wildfly.build-tools>1.1.6.Final</version.org.wildfly.build-tools> from the bom.

However the exception is from the 1.2.9 version (the latest).  So my guess would be that this is a regression with the wildfly-server-provisioning-maven-plugin - but that we shouldn't even be using the latest.

shawkins:
Specifically this is an issue with 1.2.7.Final and above.  So for now we should probably just add a fixed version to the plugin and things will work.

PR #1071, opened 2018-03-15
TEIID-5282 adding fqn and additional properties

shawkins:
Here's an initial proposal.  We can tweak the fqn format and other considerations from here.  Also rather than a single json map of extra attributes I've opted here instead for just a specific property prefix - it will be possible then to add the extension metadata attribute to each property.


rareddy:
neat and simple. I like it.

shawkins:
@mdrillin @blafond Does this make sense - you would pull the DDL and process each of the fqn's to determine the structure as seen by the source.  Additional source properties will be prefixed as teiid_rel:source_

PR #1077, opened 2018-05-02
Added quick start instructions

rkorytkowski:
One should be able to build and run teiid just looking at README.

Would you support adding a short paragraph on adding a simple vdb and issuing a query?


shawkins:
> Would you support adding a short paragraph on adding a simple vdb and issuing a query?

Yes, we can include a dummy vdb.xml or ddl in the kit.  Do you want to submit a new pull request with that or just leave this one open for now?

rkorytkowski:
I can submit a new PR.

shawkins:
Sounds good, thanks.

PR #1078, opened 2018-05-24
TEIID-4520 Support for Exasol translator

rkorytkowski:
https://issues.jboss.org/browse/TEIID-4520

### DIFF
--- a/connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/exasol/ExasolExecutionFactory.java
+++ b/connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/exasol/ExasolExecutionFactory.java
@@ -1,0 +1,405 @@
+/*
+ * Copyright Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags and
+ * the COPYRIGHT.txt file distributed with this work.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ *
+ */
+package org.teiid.translator.jdbc.exasol;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.List;
+
+import org.teiid.language.Function;
+import org.teiid.translator.SourceSystemFunctions;
+import org.teiid.translator.Translator;
+import org.teiid.translator.TranslatorException;
+import org.teiid.translator.TypeFacility.RUNTIME_CODES;
+import org.teiid.translator.jdbc.AliasModifier;
+import org.teiid.translator.jdbc.ConvertModifier;
+import org.teiid.translator.jdbc.FunctionModifier;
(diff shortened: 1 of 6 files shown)


connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/exasol/ExasolExecutionFactory.java
```diff
@@ -0,0 +1,405 @@
...
+        addPushDownFunction(EXASOL, TRUNCATE, DATE, DATE);
+        addPushDownFunction(EXASOL, TRUNCATE, DATE, DATE, STRING);
+        addPushDownFunction(EXASOL, TRUNCATE, TIMESTAMP, TIMESTAMP);
+        addPushDownFunction(EXASOL, TRUNCATE, TIMESTAMP, TIMESTAMP, STRING);
+        addPushDownFunction(EXASOL, YEARS_BETWEEN, DOUBLE, DATE, DATE);
+        addPushDownFunction(EXASOL, YEARS_BETWEEN, DOUBLE, TIMESTAMP, TIMESTAMP);
+
+        registerFunctionModifier(SourceSystemFunctions.FROM_UNIXTIME, new AliasModifier(FROM_POSIX_TIME));
+        registerFunctionModifier(SourceSystemFunctions.DAYOFMONTH, new AliasModifier(DAY));
+        registerFunctionModifier(SourceSystemFunctions.CURDATE, new AliasModifier(UPPER));
```
shawkins: This function modifier is incorrect.  Generally CURDATE would be evaluated in the engine, but this should be corrected regardless.

connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/exasol/ExasolExecutionFactory.java
```diff
@@ -0,0 +1,405 @@
...
+        registerFunctionModifier(SourceSystemFunctions.FORMATTIMESTAMP, new AliasModifier(TO_CHAR));
+        registerFunctionModifier(SourceSystemFunctions.PARSETIMESTAMP, new AliasModifier(TO_TIMESTAMP));
+
+        //Convert
+        ConvertModifier convertModifier = new ConvertModifier();
+
+        convertModifier.addTypeMapping("VARCHAR(4000)", FunctionModifier.STRING); 
+        convertModifier.addTypeMapping("DECIMAL(3)", FunctionModifier.BYTE);
+        convertModifier.addTypeMapping("DECIMAL(5)", FunctionModifier.SHORT);
+        convertModifier.addTypeMapping("DECIMAL(18)", FunctionModifier.INTEGER);
```
shawkins: Should this be DECIMAL(10) for INTEGER?

If needed we can also have the useIntegralTypes property drive the supportsConvert method to prevent integral conversions from being pushed down.

rkorytkowski: So they actually have an alias for Integer and mapped it to DECIMAL(18)... It's page 108 in their manual at https://www.exasol.com/portal/display/DOC/User+Manual+6.0?preview=/20056178/33096816/EXASOL_User_Manual-6.0.8-en.pdf

connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/exasol/ExasolExecutionFactory.java
rkorytkowski: Right!

shawkins:
We'll keep the odd integer mapping for now.  The travis test error was not related to these changes. I'll look into that.

rkorytkowski:
I've [PRed](https://github.com/teiid/teiid-documents/pull/135) docs, but not sure where to put "an entry on the overall data sources listing page"?

PR #1082, opened 2018-06-06
TEIIDTOOLS-381 changing the embedded space limit default

shawkins:


teiid-feature-pack/wildfly-integration-feature-pack/src/main/resources/subsystem-templates/teiid.xml
```diff
@@ -1,6 +1,8 @@
 <config>
     <extension-module>org.jboss.teiid</extension-module>
     <subsystem xmlns="urn:jboss:domain:teiid:1.1">            
+        <buffer-service max-buffer-space="51200"/>
```
rareddy: This is in MB instead of bytes, was that intentional?

shawkins: Yes that setting is in MB - https://github.com/teiid/teiid/blob/2ef51a5a924554f6c6b995fdb43841be4c6d004a/jboss-integration/src/main/java/org/teiid/jboss/TeiidConstants.java#L218

I see that this will need tweaked though as it matches the default, so it won't be marshaled.

rareddy: Maybe should take out the default value there, otherwise, I suspect Thorntail will pick up the value as 50GB

shawkins: My understanding is Thorntail would understand the default to be 50GB, but wouldn't pass that to TeiidAdd as it would be left unset.

I'll take out the default and see how that looks.

rareddy: The value in the DMR code it will (Thorntail), but not value in standalone.xml.

shawkins: Updated the commit.  Seems to work as expected with the default value removed from the attribute.

PR #1084, opened 2018-06-19
TEIIDTOOLS-437 adding opentracing integration

shawkins:


rareddy:
Does this mean we would need to bundle these Opentracing jars into JDBC client driver jar?

shawkins:
No, the class there is reflection loaded just like the ModuleHelper logic for local connections.  But I wanted just a single opentracing module so all of the dependencies are known to the client pom.

So for propagation to work from a remote JDBC client, you need to include opentracing-util, opentracing-api, and opentracing-noop in the driver classpath.

rareddy:
I was always thinking "Tracing" as the logging with some specific context that will be used later used to parse/filter through figure out where the service been through. Looking at your commit, it only strictly focuses on tracing part, and I see how you set up the child spans and propagate the span. What I do not understand still is how is the monitoring application, getting a wind of these events. Is it on "startSpan" and "finishSpan" the markers into the library gather these events?

rareddy:
Got it, thanks.

rareddy:
I thought you are creating a childSpan on the worker thread depending upon the type of message?

shawkins:
I added a comment yesterday about updating things to remove the in-process child-Spans.  It wasn't clear if they were needed.  Unfortunately there doesn't seem to be a consistent guide on in-process tracing of async work...  

However what I now see with jaeger makes me think that they are a good idea as references like https://www.alibabacloud.com/help/doc-detail/68035.htm makes it clear that only spans are meaningful.  

I'll update this PR with source level spans as a next step.

Another option is to make everything more request/response based, rather than only at the query execution level.

Also by using the RequestMessage to pass the span context, there is no way for this to work currently over the pg transport - but given that there's no client support there that seems ok.  At some point we may consider using a session variable to pass the context - if/when something similar is adopted by the pg community.

rareddy:
>Another option is to make everything more request/response based, rather than only at the query >execution level.
I am not sure how you see what you have now is different from request/response, as the in StatementImpl is passing whatever the application/client context down, you have an additional childSpan for Teiid. I may have to go through the above link to understand fully I guess :)

shawkins:
> I am not sure how you see what you have now is different from request/response

There's a lot more messages that the client is passing back and forth beyond just the initial query execution.  We could try to show each in its own span.

shawkins:
This is now updated to use child spans for source access.  

As of now the tracing logic is intertwined with command logging.  If command logging is not enabled then neither is tracing - we could revisit that if needed.  Beyond that there is a withActiveSpan flag on TeiidTracingUtil.  If left at false, then Teiid by itself will initiate creating traces.  If set to true, then there has to already be an active span.  We may want to default that to true, but leave it configurable via a system property.

shawkins:
Let me know if this looks good, and I'll merge it.

PR #1088, opened 2018-06-29
TEIID-5404: translator-jpa: Improve Embedded mappings handling.

Traivor:
This prevents column collisions and query failures involving Embedded
mappings by keeping track of the ManagedType traversal. Fragments of the
path are joined by '_' to form the column name. They are joined by '.'
and stored in nameInSource to be used in JPQL queries.


Traivor:
This also allows for an Enum property inside an embedded PK. I forgot to break that out into a separate commit, but can do so upon request.

rareddy:
Looks good to me

PR #1090, opened 2018-07-05
[TEIID-5270] removing truncating the staging table

vhalbert:
[TEIID-5270] removing truncating the staging table after swapping names so that current queries can complete with the state it started with


shawkins:
Should the other pr be incorporated as well: https://github.com/teiid/teiid/pull/1089

PR #1094, opened 2018-07-25
Removed superfluous dependency on opencsv

andrewruckerjones:
opencsv is no longer in use, as far as I can discern. The project builds fine without the dependency.


shawkins:
Looks good.  Seems unused to me as well.

PR #1106, opened 2018-11-05
TEIID-5524 adding minimal standalone jmx

shawkins:


shawkins:
This is a minimal standalone JMX that borrows heavily from the existing admin objects.  New bean interfaces were introduced to represent a top level Teiid mbean, nested cache mbeans, and the major stat beans (not actual mbeans).  This allows for any of the admin object / domain stuff to be left behind and to keep things to read-only attributes / operations.

We're using org.teiid for the domain, but we could just as easily use io.teiid.

It would probably be good to promote the stats beans to actual mbeans instead as that is easier for clients to deal with than CompositeData.  But it will require more code changes / wiring.

This is also always registering the beans, we could consider doing this via a system property.

rareddy:
Looks good. Yeah insted of single TeiidBean, separate beans would been little better, but I think I need to see ui montoring grouping side if that is something needed or not, I guess it is not.

PR #1110, opened 2018-12-12
TEIID-5563 pulling module dependency out of the client

shawkins:


rareddy:
Looks good, The file related ones I did not look closely. Looking back at my code the one thing I did was like

`  public interface SourceConnectionFactory<T extends SourceConnection, C extends 
    SourceConfiguration> extends Serializable {

    void configure(C config) throws SourceException;

    T getConnection() throws SourceException;

}`
to bring the configuration into it, but in hindsight I like the way you have now, as it would not be feasible to  agree on configuration in every runtime.

shawkins:
Part of the motivation of the file changes was to better isolate the VFS dependency, but I realize that is not a strong rationale as VFS is still a dependency of the engine and runtime.  

The change also does not fully remove the usage of the resource dependency - it is still required in the engine for the transactional logic.

PR #1116, opened 2019-01-07
[TEIID-5521] First, of more to come,  set of changes to moving to support the change of parent to be basepom

vhalbert:


olingo-common/pom.xml
```diff
@@ -44,12 +44,11 @@
             <groupId>org.apache.olingo</groupId>
             <artifactId>odata-commons-core</artifactId>
         </dependency>
-        
+
```
rareddy: Why are we adding a dependency, that is not correct IMO

pom.xml
```diff
@@ -1,13 +1,19 @@
...
-    <!-- <parent>
+    <parent>
       <groupId>org.jboss</groupId>
       <artifactId>jboss-parent</artifactId>
       <version>28</version>
       <relativePath />
-    </parent> -->
-    
+    </parent> 
+<!--
```
rareddy: we want basepom based Teiid, this seems wrong?? looks like someone already did basepom changes

pom.xml
```diff
@@ -23,6 +29,31 @@
...
+        <!-- Don't fork based on cores, doesn't work nicely in the cloud -->
+      
+        <basepom.check.skip-pmd>true</basepom.check.skip-pmd>
+        <basepom.check.skip-findbugs>true</basepom.check.skip-findbugs>
+        <basepom.check.skip-checkstyle>true</basepom.check.skip-checkstyle>
+        <basepom.check.skip-license>true</basepom.check.skip-license>
+        <basepom.test.timeout>600</basepom.test.timeout>
+        <basepom.git-id.skip>true</basepom.git-id.skip>
+          
+        <!--
```
rareddy: two works in progress?

rest-service/pom.xml
```diff
@@ -52,67 +88,24 @@
 			<groupId>org.jboss.spec.javax.servlet</groupId>
 			<artifactId>jboss-servlet-api_3.1_spec</artifactId>
 		</dependency>
+        <dependency>
+            <groupId>org.jboss.spec.javax.xml.bind</groupId>
+            <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
+        </dependency>
 		<dependency>
 			<groupId>io.swagger</groupId>
```
rareddy: adding and removing new dependencies for basepom change sure recipe for failure, unless these were transitive ones from before

vhalbert:
These changes were determined using the parent pom basebom.   The parent pom is not changed to basebom because the build fails with only this minimal set of changes.  That's why the complete change will be done over several commits.

PR #1120, opened 2019-01-12
TEIID-5596: Moving the Infinispan dependency as separate project to …

rareddy:
…avoid version conflicts between WildFly and Spring Boot


shawkins:
I've pushed a commit with refinements here: https://github.com/shawkins/teiid/commit/00854e959d4bc6a32225bd3fbad50d15608e67cd

- the break in the clear method causes only 1 entry be removed
- we need the localcache to be thread safe
- the embeddedconfiguration methods should just be removed rather than deprecated if they don't do anything

However since infinispan will be an optional part of the core, I'd wonder if it makes more sense to just keep the old file config methods and do a reflective creation by default - and throw an exception if the infinispan config is set, but the dependency is missing.

rareddy:
Thanks. This looks good, I forgot that we keep multiple entries under single VDB namespace.

PR #1122, opened 2019-01-13
TEIID-5596

rareddy:


shawkins:
There are still a couple of issues.  One is the new module is mpl licensed - was that intentional?  The other is that this breaks the wildfly integration.  The transactionmanagerlookup is not in the server classpath.  I'll try to address this and use reflection loading instead.

PR #1123, opened 2019-01-13
TEIID-5596 further refinements to the infinispan separation

shawkins:
Here's the follow on changes described in the other pull request.

### DIFF
--- a/cache-infinispan/pom.xml
+++ b/cache-infinispan/pom.xml
@@ -7,9 +7,8 @@
 	</parent>
 	<licenses>
         <license>
-            <name>Mozilla Public License</name>
-            <url>https://www.mozilla.org/en-US/MPL/2.0/</url>
-            <distribution>repo</distribution>
+            <name>Apache License, Version 2.0</name>
+            <url>https://www.apache.org/licenses/LICENSE-2.0</url>
         </license>
     </licenses>
     <modelVersion>4.0.0</modelVersion>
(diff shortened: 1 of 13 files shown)


cache-infinispan/pom.xml
```diff
@@ -23,6 +22,16 @@
         <groupId>org.infinispan</groupId>
         <artifactId>infinispan-core</artifactId>
     </dependency>
+
```
rareddy: Why these dependencies?

cache-infinispan/pom.xml
```diff
@@ -23,6 +22,16 @@
         <groupId>org.infinispan</groupId>
         <artifactId>infinispan-core</artifactId>
     </dependency>
+    
+    <dependency>
```
rareddy: Why these dependencies?

runtime/src/main/java/org/teiid/runtime/EmbeddedServer.java
```diff
@@ -379,15 +379,35 @@ public Object invoke(Object proxy, Method method, Object[] args)
 		}
 		
 		if (config.getCacheFactory() == null) {
-            config.setCacheFactory(new CacheFactory() {
-                @Override
-                public <K, V> Cache<K, V> get(String name) {
-                    return new LocalCache<>(name, 256);
+		    String infinispanConfig = config.getInfinispanConfigFile();
```
rareddy: I am trying to get rid of the inclusion of Infinispan directly or indirectly, put this in the hands of the developer of embedded to choose. I do not see much value in providing any reflective based impl except for testing. I feel this code needs to be in higher order frameworks like Thorntail or SB.

cache-infinispan/pom.xml
shawkins: Notice that they were moved from runtime to cache.  They are transitive dependencies of infinispan.

runtime/src/main/java/org/teiid/runtime/EmbeddedServer.java
shawkins: > I am trying to get rid of the inclusion of Infinispan directly or indirectly, put this in the hands of the developer of embedded to choose.

You are still free to set a cachefactory any way you want.  

Infinispan is completely optional with this change.  It is also backwards compatible without a code change (just a pom change) and has an appropriate deprecation.  

My position is that it's not good to silently default to a non-concurrent cache implementation, so I'd prefer this change for now and we can refine it later.

PR #1126, opened 2019-01-16
[TEIID-5594] Change to use generic dependencies

vhalbert:


connectors/misc/translator-jpa/pom.xml
```diff
@@ -39,8 +39,8 @@
             <artifactId>classmate</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jboss.spec.javax.xml.bind</groupId>
-            <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
+            <groupId>org.glassfish.jaxb</groupId>
```
shawkins: Shouldn't this be https://mvnrepository.com/artifact/javax.xml.bind/jaxb-api

connectors/odata/translator-odata4/pom.xml
```diff
@@ -54,10 +54,6 @@
 		    <groupId>org.apache.commons</groupId>
 		    <artifactId>commons-lang3</artifactId>
 		</dependency>
-		<dependency>
```
shawkins: slf4j is a transitive dependency of olingo.  Does it seem to be optional?

vhalbert: Not optional, if an exclusion is added, a ClassNotFoundException for the LoggerFactory occurs.

wildfly/connectors/webservice/connector-ws/pom.xml
```diff
@@ -144,8 +144,8 @@
             <artifactId>spring-tx</artifactId>
         </dependency>
         <dependency>
-            <groupId>org.jboss.spec.javax.xml.bind</groupId>
-            <artifactId>jboss-jaxb-api_2.3_spec</artifactId>
+            <groupId>org.glassfish.jaxb</groupId>
```
shawkins: This needs changed to the generic jaxb instead right?

connectors/odata/translator-odata4/pom.xml
shawkins: I don't quite follow the response.  It should be declared if it's actually being used - we don't want to rely on implicit transitive dependencies.

wildfly/connectors/infinispan/connector-infinispan-hotrod/pom.xml
```diff
@@ -74,6 +74,10 @@
 			<artifactId>infinispan-core</artifactId>
 			<scope>test</scope>
 		</dependency>
+		<dependency>
```
shawkins: Why is this now needed?

vhalbert: removed the dependency

wildfly/connectors/webservice/connector-ws/pom.xml
vhalbert: replaced with javax.xml.bind

PR #1127, opened 2019-01-17
TEIID-5580 upgrading to the latest mongodb driver

shawkins:


shawkins:
Tested against https://hub.docker.com/r/bitnami/mongodb latest and it connected without an issue.  Is there a good sample data set to test against?

PR #1132, opened 2019-01-30
TEIID-5620 providing a simplified set of key metrics

shawkins:


shawkins:
Here are the core metrics identified in the jira - except for the 3 pass sort count.  That is a fairly nuanced situation that we can get to later.  The only question is whether to put these on separate beans or just leave them on the root teiid bean.

PR #1133, opened 2019-02-04
TEIID-5555 adding openapi/swagger metadata

shawkins:


rareddy:
Looks good, only part I am not clear is in Spring Boot, I have redone some parts of the ODataFilter to handle for Single VDB aspect. So, what you have here may not transpire one to one, but if you can make the transformation methods public we can wire it similarly there too.

shawkins:
Here's a refinement that better isolates the handling and makes it public.  You can also control the info via a properties map.

rareddy:
Nice, thanks. I will add the SB parts.

PR #1134, opened 2019-02-07
TEIID-5521 updated van's changes for basepom alignment

shawkins:
- used targeted skip/ignore rather than explicit configuring the plugins with allowable exceptions
- sparingly removed unused / transitive dependencies - so that the poms mirror the modules. we can revisit that, which will remove a lot of the ignores
- turned off everything for dev profile
- need to re-enable checkstyle

### DIFF
--- a/connectors/couchbase/translator-couchbase/pom.xml
+++ b/connectors/couchbase/translator-couchbase/pom.xml
@@ -29,6 +29,15 @@
         <dependency>
             <groupId>org.teiid.connectors</groupId>
             <artifactId>translator-jdbc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.teiid</groupId>
+            <artifactId>teiid-admin</artifactId>
+            <scope>compile</scope>
+        </dependency> 
+        <dependency>
+            <groupId>xml-apis</groupId>
+            <artifactId>xml-apis</artifactId>
         </dependency>                 
     </dependencies>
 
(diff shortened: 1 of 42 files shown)


rareddy:
This is huge, taking a look too.

connectors/couchbase/translator-couchbase/pom.xml
```diff
@@ -29,6 +29,15 @@
         <dependency>
             <groupId>org.teiid.connectors</groupId>
             <artifactId>translator-jdbc</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.teiid</groupId>
+            <artifactId>teiid-admin</artifactId>
+            <scope>compile</scope>
```
rareddy: Do we need this scope?

olingo/pom.xml
```diff
@@ -74,6 +80,16 @@
             <groupId>org.apache.olingo</groupId>
             <artifactId>odata-server-api</artifactId>
         </dependency>
+        
+        <dependency>
+            <groupId>org.apache.olingo</groupId>
+            <artifactId>odata-commons-api</artifactId>
```
rareddy: These are just for WildFly purposes?

connectors/couchbase/translator-couchbase/pom.xml
shawkins: The dependency yes, the scope no.  That was taken directly from the branch, but it will default to compile without that of course.

rareddy:
Great! Thanks

PR #1137, opened 2019-02-25
TEIID-5656: Adding a way to filter the metadata fetch to single schem…

rareddy:
…a name in jdbc based sources

### DIFF
--- a/connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/JDBCMetadataProcessor.java
+++ b/connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/JDBCMetadataProcessor.java
@@ -88,6 +88,7 @@ public class JDBCMetadataProcessor implements MetadataProcessor<Connection>{
 	private String tableNamePattern;
 	private String catalog;
 	private String schemaPattern;	
+	private String schemaName;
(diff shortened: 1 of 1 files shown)


connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/JDBCMetadataProcessor.java
```diff
@@ -180,6 +181,16 @@ public void getConnectorMetadata(Connection conn, MetadataFactory metadataFactor
 			}
 		}
 		
+		if (this.schemaName != null) {
+			String escapeStr = metadata.getSearchStringEscape();
+			String escapedSchemaName = this.schemaName;
+			if (escapeStr != null) {
+				escapedSchemaName = escapedSchemaName.replace("_", escapeStr+"_");
+				escapedSchemaName = escapedSchemaName.replace("%", escapeStr+"%");
```
shawkins: The escapeStr will need escaped here as well.  Otherwise yes this looks good.

rareddy: ah, ofcourse. I will add that add push it. Thanks.

PR #1138, opened 2019-03-09
TEIID-5500 improving the initial delay in loading large results

shawkins:


olingo/src/main/java/org/teiid/olingo/service/LocalClient.java
```diff
@@ -236,12 +257,66 @@ public void executeSQL(Query query, List<SQLParameter> parameters,
...
         }
+        
+        int count = 0;
+        int expectedEnd = 0;
+        
+        if (!getCount && cache) {
+            //to prevent long initial requests, we want to
+            //only work over windows of at most 32 pages
+            int resultWindow = pageSize<<5;
+            int windows = 0;
```
rareddy: So what happens to data that is larger than 32 pages long? Given our default 256 rows (I think)  that is 8K rows.

olingo/src/main/java/org/teiid/olingo/service/LocalClient.java
```diff
@@ -236,12 +257,66 @@ public void executeSQL(Query query, List<SQLParameter> parameters,
...
             sql += " /* "+ sessionId +" */"; //$NON-NLS-1$ //$NON-NLS-2$
         }
         LogManager.logDetail(LogConstants.CTX_ODATA, "Teiid-Query:",sql); //$NON-NLS-1$
         
+        if (cache) {
+            this.loadingKey = Arrays.asList(sql, parameters);
+            Future<?> future = loading.get(this.loadingKey);
+            if (future != null) {
+                try {
+                    future.get(120, TimeUnit.SECONDS);
```
rareddy: So there is hard limit of 2 minutes for query execution, this is common with browser-based clients, but should we make this user configurable perhaps?

olingo/src/main/java/org/teiid/olingo/service/LocalClient.java
shawkins: Correct, but of course the agent can request a different page size.  Effectively making it unlimited by default causes a large initial delay or delay with the second page.

olingo/src/main/java/org/teiid/olingo/service/LocalClient.java
shawkins: Two minutes is almost guaranteed to see a timeout somewhere - haproxy or another gateway. SAPUI references default timeouts of 30 and 60 seconds.  Note this will only occur on the second page, or if the same user is requesting the same set of pages at the same node as is already being loaded.  It of course could be configurable.  I'll also change it to a 503 service unavailable code.

olingo/src/main/java/org/teiid/olingo/service/LocalClient.java
rareddy: I am confused, is total results are truncated to 32 pages, or first fetch is 32 pages long?

olingo/src/main/java/org/teiid/olingo/service/LocalClient.java
rareddy: then do not worry about user configuration.

olingo/src/main/java/org/teiid/olingo/service/LocalClient.java
shawkins: Fetch 32 pages at a time.   Once you need page 33, you fetch the next 32 pages.  Rather than attempting to fetch/cache the entire result set up front.  That works well until the size starts to become too large.

rareddy: So, on page 32, what is the next link you set? did I miss something here?

shawkins: All of the mechanics are still in place as before it's just that you are limiting how much of the result set you work with at one time.  That's why there are adjustments to the count/skip so that you work correctly over each window.

PR #1143, opened 2019-05-22
TEIID-5481 additional json processing functionality

shawkins:


shawkins:
Here is the simplest form of the json functionality.  There are two methods similar to xmlquery and xpathvalue.  There is also a jsonToArray that converts the json to a multidimensional array that can be processed with arraytable.  From the unit test the syntax is:

select * from arraytable(rows jsontoarray(jsonValue, '$..book[0:2]', false, '@.author', '@.title', '@.price') columns author string, title string, price decimal) as x

It's not quite as nice as having a single jsontable construct, but this can be plugged in through just a function.

Some important differences with xmltable:
- the column paths have to be relative to the context item, no parent references.
- the input document is fully parsed, no streaming processing from what I can see

PR #1144, opened 2019-05-24
[TEIID-5749] Updating libthrift versrion to 0.12.0

vhalbert:


vhalbert:
Note, this needs to be backported to the 12.2.x branch

PR #1145, opened 2019-05-29
TEIID-5481 adding json parsing functions and a jsontable construct

shawkins:


shawkins:
Here's the full initial support for the json functions and jsontable construct - which is implemented through rewrite rather than a RelationalNode.

There's still more validation that could be done - it's mostly left up to processing time.

It's also possible to use a different json provider, such as jackson-databind, rather than the minidev dependency.

Default expression support was not added we can't differentiate between null values and missing values.

PR #1150, opened 2019-06-11
TEIID-5764 defaulting to v2 and adding version parameter handling

shawkins:


olingo/src/main/java/org/teiid/olingo/web/OpenApiHandler.java
```diff
@@ -50,17 +50,18 @@
...
 public class OpenApiHandler {
 
-    enum OpenApiVersion {
-        V2("2.0", "/swagger.json"), //$NON-NLS-1$ //$NON-NLS-2$
-        V3("3.0", "/openapi.json"); //$NON-NLS-1$ //$NON-NLS-2$
+    private static final String SWAGGER_JSON = "/swagger.json"; //$NON-NLS-1$
+    private static final String OPENAPI_JSON = "/openapi.json"; //$NON-NLS-1$
 
-        private OpenApiVersion(String key, String uri) {
+    public enum OpenApiVersion {
```
rareddy: Are you sure you want to introduce our own versioning mechanism over just using OpenApi spec versioning scheme (major.minor.macro) and gleaning the version info from it?

shawkins: You mean am I worried about 2/3 vs. 2.0.0 and 3.0.0, no.  If/when minor/micro are utilized we can evaluate how we want this to be handled - at worst we'd need version parsing and select the specification that is closest, etc.  So 2/3 seems good for now.

rareddy:
Can you also merge this 12.2.1 ?

shawkins:
We're all set.

PR #1151, opened 2019-06-11
TEIID-5757 adding the odata role by default

shawkins:


shawkins:
Rather than just removing the security role, this change gives the role by default.  So this can function the same as it used to with just a config change.

The other approach would be to document the deployment overlay needed to add/remove the role.

PR #1152, opened 2019-06-14
Openapi v3 support

shawkins:


shawkins:
Here's an initial cut of the v3 source support.  It does not yet have all of the Wildfly changes to register the new translator.  It mostly piggybacks off of the existing implementation and is effectively a rewrite of the classes where the api changed.  The import properties have been simplified, such as just a url for the metadata location, and just a server, no preferred scheme - although currently if the server is not found in the server list we'll use the first server listed, we could alternatively use the configured server.

The v2 tests are mostly consistent.  There are a couple of differences:

1. There is no more BodyParameter.  Instead there can be a RequestBody defined - which I'm placing at the end of the parameters as that is consistent with the existing tests.
2. The notion of required has changed.  It seems to take into account transitivity.  So if the RequestBody is not required, but one of it's fields is, then it will still be not required.  In the old implementation, the field would still be marked as required.  See the differences in the magento ddl files.

There also seems to be an existing issue with the url formation in the old logic as there is a url https://management.azure.comnull in TestSwaggerMetadataProcessor, so I'll track that down separately.

PR #1156, opened 2019-06-17
[TEIID-5570] Upgrading apache poi to 3.17

vhalbert:


connectors/file/translator-excel/src/main/java/org/teiid/translator/excel/ExcelExecution.java
```diff
@@ -73,7 +73,7 @@ public ExcelExecution(Select query, ExecutionContext executionContext,
                 continue;
             }
 
-            Cell cell = row.getCell(index-1, Row.RETURN_BLANK_AS_NULL);
+            Cell cell = row.getCell(index-1, Row.MissingCellPolicy.CREATE_NULL_AS_BLANK);
```
shawkins: It looks like this should be Row.MissingCellPolicy.RETURN_BLANK_AS_NULL instead

shawkins:
Also https://issues.jboss.org/browse/TEIID-5570 doesn't seem to be visible.  Is that the correct jira?

vhalbert:
Not sure what's happening on accessing this jira.  I have it opened and can see it, but if I use this link that is sent in the email, it won't let me view it.  Not sure what is happening yet.

vhalbert:
Corrected the reference

PR #1161, opened 2019-06-19
[TEIID-5775] Upgrading transitive log4j to 2.9.0

vhalbert:


shawkins:
It's ok to put all of these upgrades under a single issue.  Also having them as security issues is probably unnecessary.

PR #1164, opened 2019-06-27
[TEIID-5776] Fixing CVE's related to xalan and lucene-queryparser, backported apache-poi, xerces and logging.  Also applied the teiid.patch from product build

vhalbert:


vhalbert:
Changed travis build to use the -P core-release profile

vhalbert:
The arquillian test are not going to work with the -P core-release profile and the -P release won't work because infinispan is broke with these cve fixes

vhalbert:
Updated travis build when doing arquillian-tests

vhalbert:
Updated pom.xml to remove calling wildly module so that arquillian tests will succeed.

vhalbert:
removing jdk 11 build, as its not needed at this time

PR #1165, opened 2019-07-01
[TEIID-5785] Changing profiles so that core-release only builds specific connectors

vhalbert:


shawkins:
This doesn't look right for community master. The default profile and any profile that the community (dev, final-release) uses should have all of the translators.

vhalbert:
> This doesn't look right for community master. The default profile and any profile that the community (dev, final-release) uses should have all of the translators.

Instead of duplicating all in each profile, the global modules is picked up by all profiles.  Otherwise, if u choose, I can change to remove the global modules section and then each profile will control only what it wants

vhalbert:
Changed so that all the profiles (except core-release) have all the modules.

shawkins:
What about https://github.com/teiid/teiid/pull/1165

PR #1171, opened 2019-07-17
[TEIID-5793] Upgrade commons-configuration to 1.7

vhalbert:


vhalbert:
duplicated locally with jdk11

shawkins:
When you say duplicated locally, you mean that the build fails for you with Java 11?

vhalbert:
Yes, it fails locally with java 11.

vhalbert:
I updated JDK 11 to openjdk version "11.0.4" 2019-07-16 LTS
and I no longer fail.

PR #1175, opened 2019-09-07
TEIID-5814 adding an async-load option for embedded deployment

shawkins:


rareddy:
LGTM. Couple questions
- When we did DDL based VDB, you made few VDB processing restrictions like, the definitions of all SCHEMA, should those be revisited?

- What suggestions you have for Spring Boot version as now it seems to attach a cache store for metadata easily. 

- could we even think of hooking this to database in Syndesis, that metadata is cached?

shawkins:
> When we did DDL based VDB, you made few VDB processing restrictions like, the definitions of all SCHEMA, should those be revisited?

Do you mean in relation to these changes?  I'm not sure I quite follow.  

> What suggestions you have for Spring Boot version as now it seems to attach a cache store for metadata easily.

There are a couple of different concerns and/or levels of caching at play.  The hooks that this change exposes are for by-passing parse/resolve, just like the EAP version supported.  From a container perspective you'd rather by-pass everything parse/resolve/validate (which does do metadata manipulation) - that needs to happen for the entire vdb and would require some new logic.

Also from a container/crd perspective we should convert any vdb that contains imports - of other vdbs or source metadata into a runtime image that has all metadata captured.  Otherwise there are a lot more caching and pod/cluster consistency issues that we need to deal with that we've largely ignored to this point.   Doing it this way the bulk of the image ddl would be referenced via file assets, rather than inline in the vdb definition file - https://issues.jboss.org/browse/TEIID-5745

> could we even think of hooking this to database in Syndesis, that metadata is cached?

There are really two different levels of metadata caching.  There's the base level of ddl - which we now have the mechanics for in komodo.  Then there's a higher level of caching which ideally needs to by-pass validation.  Only once we implement some kind of ttl for the dv specific preview vdbs will we need to worry about this - because otherwise they are typically reloaded only when there's a change to the dv metadata thus invalidating the cache.  Note that the syndesis logic assumes high availability of the preview / source vdbs - that logic would need to be changed to support more dynamic loading of those - and at least in it's simple form any dv preview vdb needs access to all source vdbs, see https://issues.jboss.org/browse/TEIIDTOOLS-763

PR #1177, opened 2019-09-19
TEIID-5815: correcting the salesforce usage of the extension properties

rareddy:


rareddy:
If we can back to the product branch, then we can use this in the `teiid-syndesis`

shawkins:
This means that 12.3 is broken wrt salesforce correct?

PR #1182, opened 2019-10-25
TEIID-5826: Adding ability to read length, precision and scale from m…

rareddy:
…etadata properties for OData V2, and SAP transaltors


shawkins:
An issue would be that we have some inconsistency between the notion of odata precision and teiid precision - OData timestamp precision = Teiid Timestamp scale.  See https://github.com/teiid/teiid/blob/5df9a3f228a6eb58593d20c47ed83263b853c724/olingo/src/main/java/org/teiid/olingo/service/ODataSchemaBuilder.java#L204

For the other non-numeric types precision isn't needed at all, but it should be okay to set it on them.

rareddy:
BTW, this is OData V2 only, do we know if that mismatch exists here also?

shawkins:
Yes, it looks like it: https://github.com/apache/olingo-odata2/blob/2c5a87ab54f448cfff167d95821065b529ac8952/odata2-lib/odata-core/src/main/java/org/apache/olingo/odata2/core/edm/EdmDateTime.java#L116

rareddy:
so precision on date reflects the nanoseconds on DateTime

shawkins:
> however I do not see similar behavior on OData V4, should be changed there too?

Yes that appears to have the same issue.

rareddy:
@shawkins Updated the code per your suggestions for the TIMESTAMP type.

rareddy:
@shawkins Updated the PR to left it unset for timestamp.

PR #1183, opened 2019-10-28
TEIID-5833 mostly removing the concept of namespaces

shawkins:


shawkins:
These changes don't completely remove parsing support for set namespace, rather it will fail for anything that doesn't exactly match something that is built-in.  This minimizes the breakage as we did emit the built-in declarations in our ddl.  What won't work:

* a custom prefix for a built-in namespace
* a custom namespace

The sys.properties table for now will display properties with built-in keys in both their old and new formats - note that the sysadmin.sql file did not change and it still contains references to the old uri fqn keys.  So this as well minimizes the scope of the breakage.

Codewise there is a lot of changes (constants renamed, methods removed) in public apis that should only affect those using custom namespaces.

Any new ddl to string will not emit any set namespace statements.

rareddy:
What is the procedure for defining the extension properties for the custom translator now?

shawkins:
You use whatever key you want.  We'll recommend using a prefix.  For existing users that want backwards compatibility you'd just use the uri fqn in whole.

rareddy:
ok, as long as the component and DDL match the property is valid :+1:

PR #1186, opened 2019-11-01
TEIID-5740 adding support for explain

shawkins:


shawkins:
This adds support for explain via a new command type.  In a single statement you can get the plan in text (default), xml, or yaml - and choose if you want the command to run (analyze option).  A few places need to be aware of it, but overall most logic is blissfully unaware.  It doesn't quite piggy-back on noexec as that returns no rows, so the proxy plan handles that detail.

PR #1187, opened 2019-11-01
TEIID_5837 upgrade slf4j to 1.7.28 and fix slf4j-ext so that it aligns as well

vhalbert:


vhalbert:
This fix needs to be backed ported to the 12.3.x branch.

shawkins:
merged and added to the product branch.

PR #1188, opened 2019-11-03
TEIID-5492: Upgrading Infinispan translator and connector to base on …

rareddy:
…Infinispan 10.0.1.Final


wildfly/connectors/infinispan/feature-pack-infinispan/src/main/resources/modules/system/layers/dv/org/infinispan/client/hotrod/teiid/module.xml
```diff
@@ -8,16 +8,17 @@
...
 
     <dependencies>
         <module name="javax.api"/>
-        <module name="com.google.protobuf" optional="true"/>
-        <module name="com.squareup"/>
-        <module name="org.apache.commons.pool"/>
+        <module name="com.google.code.gson"/>
+        <module name="io.netty"/>
+        <module name="io.reactivex.rxjava2.rxjava"/>
+        <module name="io.reactivex.rxjava2.rxjava"/>
```
shawkins: Looks like a duplicate line.

wildfly/connectors/infinispan/pom.xml
```diff
@@ -33,7 +33,11 @@
 				<version>${project.version}</version>
 				<type>test-jar</type>
 			</dependency>
-
+            <dependency>
+                <groupId>com.fasterxml.jackson.core</groupId>
+                <artifactId>jackson-databind</artifactId>
+                <version>2.10.0</version>
```
shawkins: Should we bump everything to 2.10?

Also I know we chatted about this, but wanted to make sure.  The databind dependency is not declared in module, only here.  Should/could it be test scoped?

wildfly/connectors/infinispan/feature-pack-infinispan/src/main/resources/modules/system/layers/dv/org/infinispan/client/hotrod/teiid/module.xml
rareddy: removed the duplicate line.

wildfly/connectors/infinispan/pom.xml
rareddy: Yes, this is test scope only. Added the scope and moved the version to a property, and updated the commit.

PR #1189, opened 2019-11-04
TEIIDTOOLS-820 making exceptions during load immediately non-recoverable

shawkins:


shawkins:
Not approved / targeted to 7.5 - there will be an issue though if there's a sql error during import.

PR #1196, opened 2019-12-07
TEIID-5835: upgrading the com.fasterxml to 2.10.1

rareddy:


shawkins:
Unfortunately this probably means there are transitive dependency upgrades needed as well.

I also locally see issues with the olingo versions of aalto-xml and joda-time - but those should be simple upgrades.

rareddy:
I wonder, why I am not seeing these dependency issues locally with on my machine :(

rareddy:
also note in wildfly we still use the WF defined version

shawkins:
Yes and this is a little broader than that.  From one of the errors something like swagger-parser-v3 2.0.12 is dependent upon jackson 2.9.  They have a 2.0.16 version that can use 2.10.  We'll need to upgrade that and anything else like that as well.  This issue would exist in teiid spring boot as well if you try to use the swagger/openapi translator.

PR #1198, opened 2019-12-17
Adding other connectors to the core-release build, minus the infinispan and sandbox

vhalbert:


vhalbert:
This needs to be back ported to the 13.0.x branch.

rareddy:
@vhalbert why are you removing this, this will force the changes in Teiid Spring Boot build depends on this. We can leave as is and not support it IMO.

vhalbert:
Not removing, but moving them to the standard build, to always be built.   Looking at the DV 7.6 doc on data sources, and the sources considered for Post GA, we need to start building all these, minus inifinispan and sandbox

PR #1199, opened 2019-12-19
Include infinispan connector in all builds, including core-release

vhalbert:
This needs to be backported to 13.0.x


rareddy:
@vhalbert this one points to master? shouldn't this be 13.x?

vhalbert:
Should be to both, because never want sandbox to be part of the prod build.

PR #1211, opened 2020-01-22
TEIID-5874 adding txn support and update the ispn connection

shawkins:


shawkins:
This aligns Teiid with the transactional changes from Teiid Spring Boot.  It goes further to refine the connection interface and it can then automatically adjust the translator properties.  The bulk logic now also defaults to putall for upsert.  I need to make an adjustment though to use infinispan-core.

shawkins:
I had to further refine things to include infinispan-core to pick up the transactionmanager.  We aren't enabling the bulk operations unless it's transactional, but that is based upon assumption rather than a full understanding of their behavior - for example they indicate that somewhere that non_xa is implemented via a synchronization, and I did not see anything about atomicity.  With the change to default to putall in the upsert case I'm effectively assuming that there are no secondary unique indexes that could be violated by such an operation, or if they are then it's more of an atomicity issue.

wildfly/connectors/infinispan/feature-pack-infinispan/src/main/resources/modules/system/layers/dv/org/jboss/teiid/resource-adapter/infinispan/hotrod/main/module.xml
```diff
@@ -19,5 +19,6 @@
         <module name="com.squareup"/> 				
         <module name="org.infinispan.client.hotrod" slot="teiid"/>
         <module name="org.infinispan.commons" slot="teiid"/>
+        <module name="org.infinispan.core" slot="teiid"/>
```
rareddy: I did not get the need for the "Infinispan-core" dependency. Is this for transaction stuff?

shawkins: Yes, it's where the wildfly transaction manager lookup is coming from.  Didn't see a great way to inject it otherwise.

rareddy: Cool, thanks for clarification.

PR #1216, opened 2020-01-23
TEIID-5883 reorganizing ispn dependencies and adding a caffeine cache

shawkins:


shawkins:
This was a little more involved than my initial thinking.  In any event this seems to completely isolate infinispan on the teiid side and move the older version to just the wildfly side.  The caffeine logic is a refinement of what is on the spring boot side to be simpler and to handle ttl - a separate issue will be needed to usage that on the spring boot side once we update the Teiid version.  To not be a breaking change we are still using the org.teiid group for the cache-infinispan.

shawkins:
> Caffeine not available in the 9.4 version of the Infinispan to use in the WildFly?

Yes it is there, but why would you want to use it directly in wildfly when you have infinispan?

rareddy:
> why would you want to use it directly in wildfly when you have infinispan?
Yes, right. Forgot that Caffine is a different cache mechanism.

PR #1219, opened 2020-01-27
TEIID-5887 changing marshalling to be thread-safe

shawkins:


shawkins:
I've expanded the test coverage slightly, but there are general issues not related to these changes captured as https://issues.redhat.com/browse/TEIID-5888

So if you are good with, we can commit this and move on to the other issues.

PR #1221, opened 2020-01-27
TEIID-5881 removing the cache aliasing

shawkins:


shawkins:
There's no mention of this feature in the formal docs, so I'll just add a release note about the removal.  In 13.0.x release notes I'll add a deprecation.

PR #1223, opened 2020-01-28
TEIID-5890 - product build fails due undeclared dependency

vhalbert:
This is a unit test build dependency errors.   The check-for-duplicate has to be turned off, due to dup's in the javax.transaction jar.


shawkins:
It would be best to not even deal with this dependency.  It is redundant with standard jta.  Can you do a dependency tree and exclude this?

vhalbert:
Unfortunately, it opened a whole different set of dependency issues when I tried excluding.   This seemed the easiest since its just a test dependency.

PR #1232, opened 2020-02-05
TEIID-5899 Update srcclr.yml and fix CVE's

vhalbert:


shawkins:
Do you need this applied forward as well?

PR #1234, opened 2020-02-05
CVE to upgrade jackson databind

vhalbert:


shawkins:
Has there ever been a release without a cve?

vhalbert:
No, because it a continual process of getting new CVE's, it seems, with every build.   And when/if dependencies are realigned, that also brings new CVE's.   We only do our best, trying to pick off the highest security CVE's (7+). 

Van Halbert

----- Original Message -----

shawkins:
Oh I meant jackson databind - it seems like as soon as one is out there's a cve.

PR #1236, opened 2020-02-13
TEIID-5904 addressing ispn bulk issues

shawkins:


shawkins:
>  What does "returnsSingleUpdateCount" capability does? I do not remember seeing that one before.

The translator api expects to return an array of update counts where each value corresponds to what happens with each row.  However many of the translators were implemented just returning a single value.  So we need to have a capability that says whether to expect the full array or just a single value.

PR #1239, opened 2020-02-21
TEIID-5909: correcting the couple issues with PG protocol when used w…

rareddy:
…ith node ja based knex library in graphback

### DIFF
--- a/runtime/src/main/java/org/teiid/odbc/ODBCServerRemoteImpl.java
+++ b/runtime/src/main/java/org/teiid/odbc/ODBCServerRemoteImpl.java
@@ -913,6 +913,10 @@ public class ODBCServerRemoteImpl implements ODBCServerRemote {
             if (modified.startsWith(PK_QUERY)) {
                 return PK_REPLACEMENT_QUERY + modified.substring(PK_QUERY.length());
             }
+
+            if (sql.equalsIgnoreCase("select version()")){ //$NON-NLS-1$
+                return "select version() as version"; //$NON-NLS-1$
+            }
         }
         else if (sql.equalsIgnoreCase("show max_identifier_length")){ //$NON-NLS-1$
             return "select 63"; //$NON-NLS-1$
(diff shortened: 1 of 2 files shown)


runtime/src/main/java/org/teiid/transport/PGCharsetConverter.java
```diff
@@ -53,6 +53,7 @@
         mapCharset("SJIS", Charset.forName("windows-932")); //$NON-NLS-1$ //$NON-NLS-2$
         mapCharset("UHC", Charset.forName("windows-949")); //$NON-NLS-1$ //$NON-NLS-2$
         mapCharset("UTF8", Charset.forName("UTF-8")); //$NON-NLS-1$ //$NON-NLS-2$
+        mapCharset("utf-8", Charset.forName("UTF-8")); //$NON-NLS-1$ //$NON-NLS-2$
```
shawkins: This change is not needed, the fallback in getCharset will already return the UTF-8 charset.

runtime/src/main/java/org/teiid/odbc/ODBCServerRemoteImpl.java
```diff
@@ -913,6 +913,10 @@ else if (sql.equals("SELECT typinput='array_in'::regproc, typtype FROM pg_catalo
             if (modified.startsWith(PK_QUERY)) {
                 return PK_REPLACEMENT_QUERY + modified.substring(PK_QUERY.length());
             }
+
+            if (sql.equalsIgnoreCase("select version()")){ //$NON-NLS-1$
+                return "select version() as version"; //$NON-NLS-1$
```
shawkins: I'd prefer not to have these kind of special purpose workarounds.  It could be an option for the resolver to auto alias in a similar manner as pg.

runtime/src/main/java/org/teiid/transport/PGCharsetConverter.java
rareddy: k

runtime/src/main/java/org/teiid/odbc/ODBCServerRemoteImpl.java
rareddy: is that a feature we need to add or already available?

shawkins: It would need to be added.  It would be a relatively small change to the auto-naming that we do to produce the exprN names.

rareddy: Ok,let's reject this PR and I will assign the JIRA to you.  I will test once you have a fix. Can we get this in 13.1 pls

PR #1244, opened 2020-02-27
TEIID-5611 adding support for scram-sha-256

shawkins:
also extracting common logic to a configuration interface


shawkins:
Once this is committed, teiid spring boot can then be simplified a little.

PR #1248, opened 2020-03-13
TEIID-5895 removing problematic fk creation

shawkins:


shawkins:
see the jira for why we should remove these fks

PR #1253, opened 2020-03-20
TEIIDSB-176 creating a common base of cxf functionality

shawkins:


shawkins:
This creates a common base for the soap logic.  It will require tweaking the dependencies on the spring boot side and disabling the security integration for now.

PR #1257, opened 2020-03-26
TEIID-5923 correcting the direct/update value handling

shawkins:


shawkins:
> So the Salesforce added native support for the time and timestamp now from string values

The problem is that the payload being created for direct operations and update added in all string values, which is not expected on the backend for some of the types.

PR #1275, opened 2020-04-14
TEIID-5927 adding support for other s3 services

shawkins:


shawkins:
This is roughly what this change will look like with using the endpoint from the connection.  It makes a couple of updates to the url handling https://docs.aws.amazon.com/AmazonS3/latest/dev/UsingBucket.html - but in general minio and ceph do not use the region based subdomains in the url.  It looks like this should work with minio just fine, but ceph only supports the older list api - https://tracker.ceph.com/issues/37801 so the v2 version won't work there.  I'll add whatever tests seem appropriate and do an end-to-end test against minio and ceph.  I don't plan on adding dockerized tests for those.

shawkins:
> I guess you could have pulled the region out of properties and make that a user-defined endpoint completely, but that may be backward compatible.

You mean pull the region out of the procedure parameters?  Yes, I would do that if not for the breaking change.

shawkins:
There was a small bug with the signing that was not including the port in the canonical request.  With that change this seems to work against minio.  I've opened a separate issue for ceph since it doesn't support the list operation.

PR #1282, opened 2020-05-01
TEIID-5916: Adding the @Field(index=Index.YES) on protobuf definition…

rareddy:
… of a Foreign table


rareddy:
Yes, I amended the commit to include support Indexes. Thanks

PR #1284, opened 2020-06-01
TEIIDSB-204 moving ws-cxf into core teiid

shawkins:


shawkins:
I had to turn off a couple of basepom checks (they were not running under wildfly) and I'll still need to validate this with Teiid Spring Boot.   There's likely a lot more cleanup and as Van was getting at we may want to go ahead and align this with the Teiid Spring Boot, Spring Boot version.

PR #1289, opened 2020-06-17
TEIID-3647 adding hdfs support to teiid wildfly

shawkins:
this further refines the dependencies of hadoop common/hdfs


shawkins:
I'll follow this up by aligning the teiid spring boot logic with this.

@aditya300899 you can see the difference in developing against wildfly with the feature pack modules changes - in particular the module.xml and ra.xml files, cli files, and additional collateral.  The nice thing is that each module gets it's own isolated classloader.  Even though we try our best to align dependencies if there is a conflict we can't resolve, this allows us to introduce multiple versions.  In particular here I'm aligning to the existing hadoop common, already used by accumulo, and matching the version of hadoop hdfs to that.  We'll need to test everything again against an actual instance to make sure this is functioning correctly.

connectors/file/file-hdfs/src/main/java/org/teiid/hdfs/HdfsConnection.java
```diff
@@ -0,0 +1,153 @@
...
+    private final FileSystem fileSystem;
+
+    public HdfsConnection(HdfsConfiguration config) throws TranslatorException {
+        this.fileSystem = createFileSystem(config.getFsUri(), config.getResourcePath());
+    }
+
+    protected FileSystem createFileSystem(String fsUri, String resourcePath) throws TranslatorException {
+        Configuration configuration = new Configuration();
+        if(resourcePath != null){
+            if (HdfsConnection.class.getResourceAsStream(resourcePath) != null) {
```
aditya300899, line 52: What is this extra check for?

wildfly/connectors/file/connector-hdfs/src/main/java/org/teiid/resource/adapter/hdfs/HdfsManagedConnectionFactory.java
```diff
@@ -0,0 +1,100 @@
...
+import org.teiid.hdfs.HdfsConfiguration;
+import org.teiid.hdfs.HdfsConnection;
+import org.teiid.resource.spi.BasicConnectionFactory;
+import org.teiid.resource.spi.BasicManagedConnectionFactory;
+import org.teiid.resource.spi.ResourceConnection;
+import org.teiid.translator.TranslatorException;
+
+public class HdfsManagedConnectionFactory extends BasicManagedConnectionFactory implements HdfsConfiguration {
+
+    private static final long serialVersionUID = -687763504336137294L;
```
aditya300899, line 33: What is this field?

aditya300899:
> We'll need to test everything again against an actual instance to make sure this is functioning correctly.

How do we do it?

connectors/file/file-hdfs/src/main/java/org/teiid/hdfs/HdfsConnection.java
shawkins: For wildfly we don't want to assume that any additional configuration resource will be on the classpath.  It's much easier for it to be at a file system path.  The existing addResource method only looks in the classpath.

wildfly/connectors/file/connector-hdfs/src/main/java/org/teiid/resource/adapter/hdfs/HdfsManagedConnectionFactory.java
shawkins: It was copied over from the ftpmanagedconnectionfactory.  Generally eclipse gives a warning is the serialVersionUID is not set on a serializable class.  We don't actually care about this concern, it's just to get rid of the warning.

shawkins:
> How do we do it?

I'll test in wildfly and I'll have you test again using your teiid spring boot test class after teiid spring boot is updated.

PR #1295, opened 2020-07-01
TEIID-5928 adding support for a partitioned load

shawkins:


shawkins:
@rareddy I'll assume that no comment is tacit approval and merge tomorrow.

PR #1297, opened 2020-07-01
dom4j fix in di-1.0-14.0.x

jrenaat:


shawkins:
This appears to be out of date.

PR #1305, opened 2020-07-03
TEIID-5936 adding s3 support to wildfly

shawkins:
this refines the tsb s3 support:
- we're using more targeted dependencies
- there's a single shared instance of the s3client as it is thread-safe
- minor code cleanups
- added an existence check for delete

some additional things that can be noted in the docs
- s3 file names report the entire name/path, unlike hdfs, ftp, or file. That is inline with the expectations of s3 though
- s3 (at least minio) cannot tolerate a leading / in names


wildfly/connectors/file/feature-pack-file/src/main/resources/modules/system/layers/dv/org/jboss/teiid/resource-adapter/s3/main/module.xml
```diff
@@ -0,0 +1,29 @@
...
+    <resources>
+        <artifact name="${org.teiid.wildfly.connectors:connector-s3}"/>
+        <artifact name="${org.teiid.connectors:file-s3}"/>
+        <artifact name="${com.amazonaws:aws-java-sdk-s3}"/>
+        <artifact name="${com.amazonaws:aws-java-sdk-core}"/>
+        <resource-root path="."/>
+        <!-- Insert resources here -->
+    </resources>
+
+    <dependencies>
```
aditya300899, line 15: What all are these dependencies?

shawkins: Modules are a separate system from poms you need to list any transitive dependencies, teiid dependencies, and even some java dependencies that are typically implicit.

connectors/file/file-s3/src/main/java/org/teiid/s3/S3Connection.java
```diff
@@ -0,0 +1,201 @@
...
+    private final AmazonS3 s3Client;
+
+    public S3Connection(S3Configuration s3Config, AmazonS3 s3Client) {
+        this.s3Config = s3Config;
+        this.s3Client = s3Client;
+    }
+
+    @Override
+    public VirtualFile[] getFiles(String s) throws TranslatorException {
+        if (s == null || s.isEmpty()) {
```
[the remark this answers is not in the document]
shawkins: Yes, that would be consistent with the file connection logic.  What I was mostly trying to correct was using the == operator with a string literal.  I'll update this.

PR #1308, opened 2020-07-06
[simpledb-api] Adding Implementation of SimpleDBConnection and Configuration along with Refactoring of Wildfly for the same.

ciphereck:
Required for Task [Add Source for Amazon SimpleDB]


ciphereck:
This is the base implementation for the common code of all Amazon Simple DB Connection.
I need to refactor a bit for better testing, but this is the basic structure I think how it's gonna go.

[Question]
The BasicConnectionFactory need to extend ResourceConnection. And for that Wildfly connector extends BasicConnection but it's depreciated.
Now since there is a cyclic dependency, I can not implement ResouceConnection in simpledb.api module.
Also, if I try to make another Implementation of SimpleDBConnectionImpl by extending SimpleDBConnectionImpl from simpledb.api, the close() method clashes as one throws Exception and other throws Resource Exception. How to proceed on that?

shawkins:
Welcome Mudit,

> Now since there is a cyclic dependency, I can not implement ResouceConnection in simpledb.api module.

In the translator or api module, you don't want any WildFly/JCA resource dependencies.  The connector/resource adapter module will be responsible for introducing that.

> the close() method clashes as one throws Exception and other throws Resource Exception. How to proceed on that?

See other examples in file-hdfs or file-s3.  The common/base connection class should ideally not throw any exception from close - which should be possible with simpledb as well.

ciphereck:
Thanks Steven.
I got it handled.
I completed refactoring of wildfly to extend the Impl of simpledb.api.SimpleDBConnectionImpl

[Question]
1. Are we going to create a single PR for the whole or maybe two PR (1. For pulling out the common code from wildfly as in this PR and 2. For Refactoring Wildfly). Second option is better accord to me as it gives better and robust way to revert the PR. Like in case of outburst.
2. I noticed in the deleteDomain(String domainName) function, we are removing the domain from list of domains, but we are not adding domain again in list of domain when someone calls createDomain(String domainName). I'm little confused here.
3. Also, in the SimpleDBManagedConnectionFactory, should we store SimpleDBConfiguration Object instead of accessKey and secretAccessKey? I'm not sure how these are getting set. That's why the confusion.

Let me know your thoughts on this.

ciphereck:
Also @shawkins @rareddy 
Can you please review my commits and see if everything is all right?

shawkins:
> Second option is better accord to me as it gives better and robust way to revert the PR. Like in case of outburst.

I am fine with either.  Master is only picked up in snapshots and is understood to have some instability until a release.

> I noticed in the deleteDomain(String domainName) function, we are removing the domain from list of domains, but we are not adding domain again in list of domain when someone calls createDomain(String domainName). I'm little confused here.

It looks like this was not implemented completely.  It also may not be what we want.  The "delete from tbl" case is being implemented more as a drop rather than a delete.  Then the insert case is allowing for the implicit creation of a domain.  This does not match our behavior with most sources, such as traditional databases.  In general we expect the source structure to exist with some additional controls over auto-creation (such as with the infinispan source).  For now I'd just remove the usage of the domains member variable.

> Also, in the SimpleDBManagedConnectionFactory, should we store SimpleDBConfiguration Object instead of accessKey and secretAccessKey? I'm not sure how these are getting set. That's why the confusion.

The ManagedConnectionFactory typically will be an instance of the configuration interface - the JCA framework treats the methods on that class as getters/setters for configuration defined by the associated ra.xml file.  So you don't need to include a base implementation of the configuration class in the api module.  See https://github.com/teiid/teiid/blob/master/wildfly/connectors/file/connector-s3/src/main/java/org/teiid/resource/adapter/s3/S3ManagedConnectionFactory.java

The builder class you have associated with SimpleDBConnectionImpl serves the same purpose have called a ConnectionFactory - https://github.com/teiid/teiid/blob/master/connectors/file/file-s3/src/main/java/org/teiid/s3/S3ConnectionFactory.java such that any validation and common state can be held there and reused in Teiid Spring Boot https://github.com/teiid/teiid-spring-boot/blob/master/data/s3/src/main/java/org/teiid/spring/data/s3/S3ConnectionFactory.java

ciphereck:
@shawkins Thanks for the insight. I've updated code accord to your comments.
Can you please review now?

shawkins:
> Can you please review now?

I think this looks good, but we should still keep the unit test as well.

ciphereck:
@shawkins @rareddy 
[Question]
How to test a public method that calls another private method? (ex. getAttributeNames(String domainName) in SimpleDBImpl class)
[Possible Sol.]
1. Make the method public (Doesn't make much sense)
2. Just call the public method in test class with such argument, so that it doesn't fail (ex return 0 while calling attributeNameCount from metaDataResult) (Works, I think it stills leaves some cases behind)
3. If some way we can check whether a particular method is getting called or not with specific arguments, that will be best. (But since private, it's not visible in TestFile) 

Can you tell me how to proceed? I checked on internet, @TestVisible annotation, or PowerMock library might be a good lead.

ciphereck:
I have pushed my works so far on Unit Tests. Can you check once?

[Question]
1. Do I need to add unit tests for private methods too? If yes, then how to make it visible?

[Help]
1. Some of the method like performInsert or performUpdate are a little big. Can you tell me how to add unit tests for those. I couldn't find any unit tests written previously on the same.

shawkins:
> Do I need to add unit tests for private methods too? If yes, then how to make it visible?

In general no.  If you can test through the other public methods, that is for the best.  If it seems important to increase coverage, then you can change the method to package private or protected to make it more easily testable with junit/mockito.

> Some of the method like performInsert or performUpdate are a little big. Can you tell me how to add unit tests for those. I couldn't find any unit tests written previously on the same.

My primary concern was for the original coverage to be preserved.  If you want to add additional coverage that would be great.  For the larger methods you can still do something similar of stubbing out / verifying the expected behavior on your mock - which may be quite a bit of setup unfortunately.  Using verify https://www.baeldung.com/mockito-verify in simple tests like your create or delete domain would be quite straight-forward.

ciphereck:
@shawkins 
I have updated createDomain and deleteDomain with verifying the invocation on mock.
Also, previous coverage is preserved, as well as I added some more tests for the same.
I think if everything looks good to you then maybe we can move towards final review?

shawkins:
Thanks Mudit, I think the code looks good now.  The final thought is to simplify the module structure - there isn't much benefit to having a separate api module when it leaks the aws-java-sdk-simpledb dependency.  Can you combine the simpledb translator and api maven modules?

The only nuances to that are removing the simpledb-api declared dependency from the root pom and updating the wildfly modules https://github.com/teiid/teiid/tree/master/wildfly/connectors/amazon/feature-pack-amazon/src/main/resources/modules/system/layers/dv/org/jboss/teiid

ciphereck:
@shawkins 
I simplified the module structure by removing simpledb-api module and merging it inside translator-simpledb module.
Also changed the dependency of wildfly-simpledb. Can you check once, if everything is removed regarding simpledb-api is removed or not?

If everything is removed and the simplification of module was done properly accord to you and if build passed, I think we can go ahead with closing the PR.

wildfly/connectors/amazon/feature-pack-amazon/src/main/resources/modules/system/layers/dv/org/jboss/teiid/translator/simpledb/main/module.xml
```diff
@@ -14,6 +14,6 @@
         <module name="javax.resource.api" />
         <module name="org.jboss.teiid.common-core" />
         <module name="org.jboss.teiid.api" />
-        <module name="org.jboss.teiid.translator.simpledb.api" />
+        <module name="org.jboss.teiid.translator.simpledb" />
```
shawkins, line 17: This line can be removed as it's now just declaring a dependency on itself.

ciphereck: I'll update it first thing in the morning. 
Turned the laptop off just now.

shawkins:
There should have been an IntegrationTestDeployment smoke test of creating a datasource on wildfly for simpledb, similar to the others: https://github.com/teiid/teiid/blob/7f97cbf41f202d6e555c128ea605798c36dea0c7/wildfly/test-integration/common/src/test/java/org/teiid/arquillian/IntegrationTestDeployment.java#L729

I added one and it fails because the hashcode and equals methods were removed from the SimpleDBManagedConnectionFactory:

Caused by: org.jboss.jca.deployers.common.DeployException: IJ020056: Deployment failed: file:/home/shawkins/workspaces/teiid-git/teiid/wildfly/test-integration/common/target/teiid-15.0.0-SNAPSHOT/modules/system/layers/dv/org/jboss/teiid/resource-adapter/simpledb/main/                                          
    Caused by: org.jboss.jca.validator.ValidatorException: Severity: ERROR                                                                                   
Section: 6.5.3.2                                                                                                                                             
Description: A ManagedConnectionFactory must implement a \"public boolean equals(Object)\" method.

It's kind of a odd restriction in that it can be satisfied just by calling the default methods, but that will also be needed.

ciphereck:
@shawkins 
I pushed the smoke test in integration test module. But the build is failing. Can you help please?

shawkins:
Further back in the log is:

Caused by: java.lang.NoClassDefFoundError: com/amazonaws/auth/AWSCredentials
	at org.teiid.resource.adapter.simpledb.SimpleDBManagedConnectionFactory.createConnectionFactory(SimpleDBManagedConnectionFactory.java:52)
	at org.teiid.resource.spi.BasicManagedConnectionFactory.createConnectionFactory(BasicManagedConnectionFactory.java:53)
	at org.jboss.jca.deployers.common.AbstractResourceAdapterDeployer.createObjectsAndInjectValue(AbstractResourceAdapterDeployer.java:1788)
	... 10 more

This means some more refinement of the wildfly modules is in order.  There is a com.amazonaws module that is now not being referenced, which declares the dependencies on aws-java-sdk-core and simpledb.  If you want, I can do that refinement and create a common module for aws-java-sdk-core and then commit this so that we can have you move on to the spring boot side of things.

ciphereck:
Can you tell me what kind of refinement are we talking about?

shawkins:
This is what I came up with: https://github.com/shawkins/teiid/commit/49a0b8a5827bbe757b2bd616b532cf5f467e28e3

That creates a shared amazonaws module and will allow for the creation of the connection factory even if a client cannot be created at that time.

ciphereck:
1. What's the use of AWSStaticCredentialsProvider
Like I like to know the tradeoff about why we used it?
2. Why the connection factory is not able to create the client?
3. How shared amazonaws module helps to resolve the issue? (I got the singleton use of SimpleDB, but didn't get the use of amazonaws module)

Also, I think I added a type getS3Client() in SimpleDBConnectionFactory class.
If everything else seems fine to you, maybe we can merge the two commits and then merge it in the master if build passes?

wildfly/connectors/file/feature-pack-file/src/main/resources/modules/system/layers/dv/org/jboss/teiid/resource-adapter/s3/main/module.xml
```diff
@@ -25,5 +24,6 @@
         <module name="com.fasterxml.jackson.core.jackson-annotations"/>
         <module name="org.apache.httpcomponents.core"/>
         <module name="org.joda.time"/>
+        <module name="com.amazonaws" />
```
ciphereck, line 27: Why this change?
Like in module.xml from s3?

shawkins:
> Like I like to know the tradeoff about why we used it?

That's the only credentials type that is currently supported.  It's used to pass the secret/access key.

> Why the connection factory is not able to create the client? If it's not able to create the client, isn't it's gonna mess with further functionalities? Like when in spring-boot perform some action on the client? (it might be performing it on the null client reference? since we are not throwing the exception outside the class?)

It will never be null.  Either you get the client or you get an exception.  If we throw the exception at construction time that prevents the datasource from being created in WildFly - but the exception may be transient such as a network issue in reaching the simpledb instance.

> How shared amazonaws module helps to resolve the issue? (I got the singleton use of SimpleDB, but didn't get the use of amazonaws module). And what is the meaning of the shared amazonaws module? I didn't get that?

It's the same dependency used in multiple places.  By creating a shared module that dependency is included only once in the kit and we can better manage the transitive dependencies, those that are marked for export.

> And getSimpleDBClient is not throwing any exception, that what SdkClientException exception are we catching?

It does not throw any checked exceptions, but it does throw RuntimeExceptions, such as SdkClientException which is their root exception.

wildfly/connectors/file/feature-pack-file/src/main/resources/modules/system/layers/dv/org/jboss/teiid/resource-adapter/s3/main/module.xml
shawkins: The module for com.amazonaws:aws-java-sdk-core is called com.amazonaws (that was the existing module name, but it would be made clearer).  And yes it is used by both simpledb and s3.

shawkins:
> Also, I think I added a typo getS3Client() in SimpleDBConnectionFactory class. I have updated it. So need to worry about it.
If everything else seems fine to you, maybe we can merge the two commits and then merge it in the master if build passes?

Yes, I believe that everything should be good now.  I'll look things over once again and then merge everything.

ciphereck:
> That's the only credentials type that is currently supported. It's used to pass the secret/access key.

No I mean instead of BasicAWSCredential why we are using  AWSStaticCredentialsProvider

> It will never be null. Either you get the client or you get an exception. If we throw the exception at construction time that prevents the datasource from being created in WildFly - but the exception may be transient such as a network issue in reaching the simpledb instance.

 Okat got that part. But if ConnectionFactory gets created even if the client cannot be created at that time (with exception), then the reference of simpleDBClient is going to be null accord to me. Isn't it? Like when we are logging the error.

ciphereck:
I see you merged the PR. I'll now start working on spring-boot side of things.

shawkins:
> No I mean instead of BasicAWSCredential why we are using AWSStaticCredentialsProvider

AWSStaticCredentialsProvider is just wrapping the BasicAWSCredential and an instance of AWSCredentialsProvider is what is expected by the builder.

> But if ConnectionFactory gets created even if the client cannot be created at that time (with exception), then the reference of simpleDBClient is going to be null accord to me. Isn't it? Like when we are logging the error.

Yes it will be null, but the getter will attempt to create it if it is null.

PR #1313, opened 2020-07-18
Lazy validation

3DRaven:
Every time connection validation changed to lazy. Because make call
{code}
partnerConnection.getServerTimestamp()
{code}
every inbound request it is very expensive and give low performance.
Connection "is alive" now will fired if we get Exceptions.

### DIFF
--- a/connectors/salesforce/translator-salesforce/src/main/java/org/teiid/salesforce/BaseSalesforceConnection.java
+++ b/connectors/salesforce/translator-salesforce/src/main/java/org/teiid/salesforce/BaseSalesforceConnection.java
@@ -160,16 +160,12 @@ public abstract class BaseSalesforceConnection<T extends SalesforceConfiguration
             s = s.substring(index+1, end);
             return Long.valueOf(s);
         } catch (NumberFormatException e) {
-            checkValid();
             throw new TranslatorException(e);
         } catch (MalformedURLException e) {
-            checkValid();
             throw new TranslatorException(e);
         } catch (UnsupportedEncodingException e) {
-            checkValid();
             throw new TranslatorException(e);
         } catch (IOException e) {
-            checkValid();
             throw new TranslatorException(e);
         } finally {
             if (is != null) {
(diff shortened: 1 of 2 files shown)


ciphereck:
@3DRaven I see that you are creating PR again for the same thing, maybe you can directly push commit into your corresponding branch? It will automatically update the PR and fire the build.

3DRaven:
A, ok :) In future I will don't create new PR for every fix :) Tnx.

3DRaven:
I just don't know rules in this project, so I thought need new PR every fix :)

ciphereck:
- You can create a PR for every different fix.
- I don't think it's a rule, but it's like followed convention. If you created a PR for fixing A, and want more code to add, add code in the same branch. 
- Also, for every new file created, you have to add the licence, which you can find in every file of the project. Now, this is a rule and if not followed, the build will not pass on that PR.
- Let me know if I can be any more of help.

shawkins:
Generally this type of change is not needed.  You can configure on the pool the validation strategy you want:

https://docs.wildfly.org/19/wildscribe/subsystem/resource-adapters/resource-adapter/connection-definitions/index.html

By default test on lease is used, but you can also use background validation.  See also the flush strategy.

Also it would also be cleaner to approach something this with a proxy, but I'm not sure it's needed at least in base teiid.

Can you explain why getUserInfo is a better call to make?  The server timestamp result payload is much smaller and it would not seem to be an expensive call.

3DRaven:
`Can you explain why getUserInfo is a better call to make?`

We have some problems with getServerTimestamp where this call worked in invalid connection. We do not have any information "Why", sorry.

`Generally this type of change is not needed. You can configure on the pool the validation strategy you want:`

This functionality is not working for teiid-spring-boot.

shawkins:
> We do not have any information "Why", sorry.

If you can find a reference to or an example of the problematic behavior, we can make that change.  Otherwise we'll need to leave it at the current call.

> This functionality is not working for teiid-spring-boot.

I understand.   What I'm hoping is that this and your other Teiid pr could be satisfied by adding pooling functionality for non-jdbc sources to Teiid Spring Boot.

3DRaven:
Ok, because in my other PR we have "Ok, poll" i close this PR :)

PR #1317, opened 2020-07-28
DynamoDB Connector

ciphereck:
Signed-off-by: Mudit Jain <ciphereck@gmail.com>


connectors/amazon/translator-dynamodb/src/main/java/org/teiid/translator/dynamodb/DynamoDBConnectionFactory.java
```diff
@@ -0,0 +1,50 @@
...
+        awsStaticCredentialsProvider = new AWSStaticCredentialsProvider(credentials);
+
+
+    }
+
+    public AmazonDynamoDB getDynamoDBClient() {
+        if (dynamoDBClient == null) {
+            synchronized (this) {
+                if (dynamoDBClient == null) {
+                    dynamoDBClient = AmazonDynamoDBClient
```
ciphereck, line 37: Here I have a question. We can go two ways either with AmazonDynamoDBClient or with AmazonDynamoDBAsyncClient.

The AmazonDynamoDBAsyncClient follows async programming as the name suggests.
Ref: https://docs.aws.amazon.com/sdk-for-java/v2/developer-guide/basics-async.html

Let me know, which way to move forward.

ciphereck:
Also, as I figured every translator has plugin, that might be of use to turn on / off a particular feature?
And enum Event, which maps to specific failure, I think? And it is use to track where the server is misbehaving, in case of outage? Is that correct?

As, I see Accumulo events enum starts with 19k's and SimpleDB starts with 24k's. What is going to be event numbers for DynamoDB? And how we decide that? Is it just increasing number from the last developed source?

ciphereck:
Also, for moving forward, I'm guessing we need to add some methods in DynamoDB Connection and implement them?

Can you give me a little roadmap here? Maybe just how control is gonna flow in the files and then I'll try to figure the code out? For just one query?

cc: @shawkins

shawkins:
> Also, as I figured every translator has plugin, that might be of use to turn on / off a particular feature?

No it's a naming convention hold over from the days when MetaMatrix was developed using the eclipse plug-in framework.  Feature enablement for translators is usually controlled via execution properties that have set methods on the ExecutionFactory class.

> And enum Event, which maps to specific failure, I think? And it is use to track where the server is misbehaving, in case of outage? Is that correct?

Yes it is to provide a unique code and allows for other language files to be developed (in practice all we maintain in the project is the English language translation).

> What is going to be event numbers for DynamoDB? And how we decide that? Is it just increasing number from the last developed source?

Yes, just start with the next available which seems to be TEIID32xxx

> Also, for moving forward, I'm guessing we need to add some methods in DynamoDB Connection and implement them?

You are starting with a port of the existing logic from simpledb to dynamodb - swap the dependencies and correct the code as much as possible to the dynamodb api.  Once it becomes clear how the logic needs to evolve, then yes you are free to change/add methods to the connection.

connectors/amazon/translator-dynamodb/src/main/java/org/teiid/translator/dynamodb/DynamoDBConnectionFactory.java
shawkins: Start with the sync version as that matches what was done in simpledb.

You can look at doing the async later if you want once you have a better handle on things.  We have a simple async programming model.  See the DataNotAvailableException and it's usage in the Salesforce, Cassandra, or loopback translators along with http://teiid.github.io/teiid-documents/master/content/dev/Executing_Commands.html

Essentially if the engine asks for data from the translator and it does not have it, then you can have the translator throw an exception that may indicate a polling value.  If data arrives sooner than the polling value, the translator can notify the engine via ExecutionContext.dataAvailable.

ciphereck:
I added some DDL methods in dynamodb connection. I'll add more in some time.

As per the translator, this is what I got so far.
- ExecutionFactory - Primary Entry point for translator. Shows capability of translator. Use to return metadata or connection etc. Also use for execution of various queries.
- QueryExecution - Reponsible for executing a query.
- DeleteExecute, UpdateExecute - Use to perform a query based on name with the help of connection, where method from connection is called. Also, it uses visitor to get atom size component/node of query.

This is what I got so far. Is that right?

Also, I'm confused with below:
- MetadataProcessor - Ik it's used to process metadata, but what kind of metadata are we talking about? Like can you give me an example?
- DirectQueryExecution - Not sure, what is the use for this. I see it's mentioned that it's used for execution of call? But what is the meaning of procedural call?

Also, it there some documentation or any doc for connectors (or any other info in a document)? That might be of great help.

Also, on the same discussion, I don't think there's an issue on Jira Dashboard for this. Do we need to create one?

cc: @shawkins

shawkins:
> Is that right?

Yes, that is a good description.

> MetadataProcessor - Ik it's used to process metadata, but what kind of metadata are we talking about? Like can you give me an example?

It is meant to process the external source metadata into Teiid source metadata, ultimately calling methods on the MetadataFactory to add Tables, Procedures, Functions, etc.  That metadata conversion will be different for every source - they will have different source representations (MongoDB collections, Salesforce Entities, Oracle Tables, etc.), different data types that are supported, etc.  We just need the best representation in Teiid that the translator will actually support querying.  The SimpleDB logic looks quite simple - for each Domain add a table with a primary key.  There are quite a few other translators to look at, but most are more complicated than that.

> DirectQueryExecution - Not sure, what is the use for this. I see it's mentioned that it's used for execution of call? But what is the meaning of procedural call?

It is a source procedure to by-pass Teiid to directly issue a source query.  Let's say you have a source that supports it's own query language like salesforce SOQL.  If you access salesforce through Teiid tables, then the translator will generate the SOQL from the pushdown query.  If for whatever reason you need more control, then you can enable the direct query procedure so that you can pass SOQL directly - call salesforce_source.native('some really specific soql').  See the original issue for SimpleDB as well https://issues.redhat.com/browse/TEIID-2760 

 > Also, it there some documentation or any doc for connectors (or any other info in a document)? That might be of great help.

The docs we have are at http://teiid.github.io/teiid-documents/master/content/dev/Developing_JEE_Connectors.html

WildFly/EAP has much more https://access.redhat.com/documentation/en-us/jboss_enterprise_application_platform/6/html/administration_and_configuration_guide/chap-java_connector_architecture_jca

> Also, on the same discussion, I don't think there's an issue on Jira Dashboard for this. Do we need to create one?

Yes, I don't see an existing issue.  Can you create one?

ciphereck:
https://issues.redhat.com/browse/TEIID-6003

ciphereck:
I added Delete Execution to delete table. I'm still in completing that for deleting a specific record. It's copied mostly from Simpledb, but still can you have a quick look?

**Questions**
- criterial is where clause right? If there is no where cause, we are assuming user wants to delete the whole table?
- If we are deleting table with "delete from tableName". What about the DDL command "drop table tableName"?
- What is parameter values in Delete command class? Intuition tells me it's the parameter on which data is going to delete like delete row whose studentName='Mudit'. Is that correct? Also, if parameterValues are not null, we are adding exception? Why?
- Why we are throwing the first exception in visitor everywhere? Is there some logic. Btw I think there are only going to be one exception. And if there are multiple threads executing the same execute. can we make the list thread safe?
- I was thinking of writing unit test for checking if everything is working fine or not? I'm confused with the line `TransformationMetadata tm = RealMetadataFactory.fromDDL("create foreign table item (\"itemName()\" string, attribute string, somedate timestamp, strarray string[]);", "x", "y");` I see this is used many places for parsing the query. Can you explain this?

**Update**
- Instead of a String criteria which simply have sql query in string if there is where clause, I'm storing the where condition itself. I hope it's okay. Or you want me to have a boolean variable for the same? I think boolean will more make sense. Thoughts? 
- I added some code from simpledb directly. I'm still not sure, how it's working. Although I'm still trying to figure them out. Mostly, it's visit node code.

cc: @shawkins

ciphereck:
I was trying to move for my next InsertExecute. I noticed createDomain function was not used in any of the translator file. The only function using createDomain function from SimpleDBConnection is in the test. No where else. 
I'm not sure if there is no use of createDomain in translator, why this function is there. Or it got left out?
Can you check and let me know on this?

shawkins:
> criterial is where clause right? If there is no where cause, we are assuming user wants to delete the whole table?

Yes

> If we are deleting table with "delete from tableName". What about the DDL command "drop table tableName"?

No that is not the same.  delete from tableName - semantically means remove all records in an atomic delete that can be rolledback.  drop table tableName - is a ddl operation that typically cannot be rolled back, often destroys associated constructs like indexes, and may not be expected by other clients in that the structure will no longer exists.  Systems like databases have another related statement type - truncate https://www.tutorialspoint.com/difference-between-delete-and-truncate-in-sql-query#:~:text=Delete%20and%20truncate%20both%20commands,for%20deleting%20the%20specific%20data that could be used but could need a hint or execution property to control.

> What is parameter values in Delete command class? Intuition tells me it's the parameter on which data is going to delete like delete row whose studentName='Mudit'. Is that correct? Also, if parameterValues are not null, we are adding exception? Why?

See the docs on bulk and batch commands.  If there are parameter values would only be from a JDBC prepared statement - "delete ... where studentName = ?" - and the user adds more than 1 student name to a single execution.  The ExecutionFactory must return supportsBulkUpdate=true for those values to be pushed.  Bulk insert is much more common, which looks to be what the initial implementation was trying to support.

> Why we are throwing the first exception in visitor everywhere? Is there some logic. Btw I think there are only going to be one exception. And if there are multiple threads executing the same execute. can we make the list thread safe?

The problem is with the visitor visit methods - they do not throw a checked exception, so you either need to save the checked exception or throw the exception as unchecked and possibly unwrap it somewhere else.  There is no threading concern as the usage of a visitor is single threaded.

> I was thinking of writing unit test for checking if everything is working fine or not? I'm confused with the line TransformationMetadata tm = RealMetadataFactory.fromDDL("create foreign table item (\"itemName()\" string, attribute string, somedate timestamp, strarray string[]);", "x", "y"); I see this is used many places for parsing the query. Can you explain this?

It creates our main internal class for accessing metadata information in the engine - TransformationMetadata - from a ddl string.  That is sufficient for a lot testing - no need to create a full vdb, start a full server, etc.

> I'm not sure if there is no use of createDomain in translator, why this function is there. Or it got left out?

The refactoring that you did removed the usage of that method.  Previously it allowed for the lazy creation of a domain based upon the first insert.  In general that is not something that matches our other translator usage patterns - we don't issue create/drop, we expect the structures to already exist.  While there's some ease of use value to this often times there is not complete metadata as to how to create the source structure - such as missing indexing and other attributes not captured in Teiid source DDL.  At the very least this type of behavior needs to be driven by a execution property.  I don't think you should account for create/drop in your initial implementation.

shawkins:
https://github.com/teiid/teiid/commit/b3990331959579536585066eaaecae8c2e61cd35 makes thing symmetrical with simpledb - create/delete domain are no longer used by the tranlator.

ciphereck:
>No that is not the same. delete from tableName - semantically means remove all records in an atomic delete that can be rolledback. drop table tableName - is a ddl operation that typically cannot be rolled back, often destroys associated constructs like indexes, and may not be expected by other clients in that the structure will no longer exists. Systems like databases have another related statement type - truncate https://www.tutorialspoint.com/difference-between-delete-and-truncate-in-sql-query#:~:text=Delete%20and%20truncate%20both%20commands,for%20deleting%20the%20specific%20data that could be used but could need a hint or execution property to control.

Yes, Agree. But in simpledb when user is trying to execute delete from tableName without a where clause, instead of deleting data from domain (truncate as you mentioned), we are actually deleting domain in that case. Is there a catch on this?

[Update:]
After you mentioned https://github.com/teiid/teiid/commit/b3990331959579536585066eaaecae8c2e61cd35 it makes a lot of sense now. It answers my question itself.

> >Why we are throwing the first exception in visitor everywhere? Is there some logic.

>The problem is with the visitor visit methods - they do not throw a checked exception, so you either need to save the checked exception or throw the exception as unchecked and possibly unwrap it somewhere else. There is no threading concern as the usage of a visitor is single threaded.

Okay. Got it. But my question was why we are throwing the unwrapping the first occurred exception? Why not the last one? For me, the last one make sense, because, that is the latest occurred exception. That might be more relevant compared to the first occurred one?

>The refactoring that you did removed the usage of that method. Previously it allowed for the lazy creation of a domain based upon the first insert. In general that is not something that matches our other translator usage patterns - we don't issue create/drop, we expect the structures to already exist. While there's some ease of use value to this often times there is not complete metadata as to how to create the source structure - such as missing indexing and other attributes not captured in Teiid source DDL. At the very least this type of behavior needs to be driven by a execution property. I don't think you should account for create/drop in your initial implementation.

Okay. Got it. Thanks for the detailed explanation. This is lots of help. Instead of focussing on creating structure using createTable, I'll now focus more on performing Update or insert from now on. Assuming the structure is already there. Thanks. Great Help.

> It creates our main internal class for accessing metadata information in the engine - TransformationMetadata - from a ddl string. That is sufficient for a lot testing - no need to create a full vdb, start a full server, etc.

I'll visit the code once more and see if your follow back makes more sense to me. Otherwise, I'll ask/raise question again on this.

These comments were of great help. And I got some direction again to go with. Also, thanks for the quick fix on deleteDomain.

shawkins:
> Okay. Got it. But my question was why we are throwing the unwrapping the first occurred exception? Why not the last one? For me, the last one make sense, because, that is the latest occurred exception. That might be more relevant compared to the first occurred one?

Typically if there's an exceptional situation everything that comes after that in the visitation is invalid.  I generally prefer the pattern of throwing a runtime exception (usually TeiidRuntimeException) that can wrap a checked exception and let the engine handle it from there - https://github.com/teiid/teiid/blob/1ec2db1c0b19a177fb108b501d34a4602f9d6802/engine/src/main/java/org/teiid/dqp/internal/datamgr/ConnectorWorkItem.java#L322

ciphereck:
Got it now. Works for me.

As per my forced commit, just rebased my branch from the latest master. (This is just for the records in the PR)

ciphereck:
So, I got a blocker in performSelect.

Basically there are four ways to get items from the table.
1. batchGetItem
2. getItem
3. query
4. scan

In simpledb, we executed the sql query directly using select expression. But as far as I could find, dynamodb doesn't support that.

Also, batch update is not supported in dynamodb. We can only update only one item at a time. For deletion or updation, we can do it in batch of 25.

Now, obviously I think we are not going to use getItem, since it only returns one row. And it will increase a lot of cost. The same for scan, cause it reads the whole table.

What I need is to decide whether to go with batchGetItem or query. Both of them extract data based on different parameters. Can you help me to deice what exactly to go with, when performingSelect?

cc: @shawkins 
ref: https://docs.aws.amazon.com/AWSJavaSDK/latest/javadoc/com/amazonaws/services/dynamodbv2/AmazonDynamoDBClient.html

shawkins:
> What I need is to decide whether to go with batchGetItem or query. Both of them extract data based on different parameters. Can you help me to deice what exactly to go with, when performingSelect?

getItem - you would use this when a single primary key value is available.  In the pushdown that would look like "where pkcol = 'value'"

getBatchItem - you would use this with multiple primary key values.  In the pushdown that would like "pkcol IN ('value', 'value2', ...)"

query - seems to a have a specific usage pattern.  It requires the item has a composite key where part of the key is used for partitioning and that a partitioning value must be specified in the pushdown query.  There then are additional key searches and filtering that can be done in the query request based upon the remaining columns.  To use this you will at least need additional metadata about what column of the primary key is the partitioning column (I don't see at a first glance if multiple partitioninings are supported).

scan - this is very general and yes would be the most expensive.  If no key (primary or index) information is available, then this is best you can do.  In the pushdown query this could be something like "where description like '%foo%'".

In an initial version of the translator you want to focus on only a couple of these scenarios that would be fine - for example require a primary key, only support equality predicates, etc.  You can then expand support from there.

ciphereck:
I thought of going with batchGetItem in the initial steps for insert command. I tried making a batchGetItemRequest object for a simple query if I have a tablename and other info, but I couldn't able to make it due to very low level api stuff.

Tried and got blocked by it. I tried contacting some known employee of Amazon who works a lot with dynamodb. He suggested me to use Dynamo DB Mapper for read, update or delete operations.
 https://docs.aws.amazon.com/amazondynamodb/latest/developerguide/DynamoDBMapper.html

This is developed by AWS only. Since, I couldn't move ahead with low level api directly, I'm thinking of moving ahead with this as suggested by someone. 

Can you tell me what do you think about it? 
I just got recommendation of dynamo db mapper, so couldn't got a chance to look at it, but as soon as I look into it, I'll update more here.

I just have a look at the initial docs, and found out that we can perform CRUD with table data, but not with table exactly, which is our use case any how I think for now?

rareddy:
Looks like DynamoDBMapper depends on the Java annotations for code execution, so I do not think it will be possible to use that!

ciphereck:
Just wondering, why we can not use java annotations in teiid? @rareddy I think I'm missing something?

rareddy:
When/if you write your data model as POJOs then you have the ability to add annotations on that model, and based on those POJOs you can use the API. But in the Teiid case, there is no POJOs or code generation for it, the model/schema is being constructed dynamically by looking at the DynamoDB's metadata, so you will be able to use it

PR #1319, opened 2020-07-30
TEIID-5780 adding authentication based upon ssl mutual auth

shawkins:


shawkins:
> This would be very useful with our cloud model now.

Right, I had started this as part of the effort for tight integration with pg such that certs could be used to authenticate.  I've updated to have the same vdb property handling as gss and password as well.

PR #1320, opened 2020-08-06
TEIID-5980 upgrade postgresql driver to 42.2.13

vhalbert:


shawkins:
Incorporated into another set of changes.

PR #1322, opened 2020-08-13
Add support for multiple asterisk in glob search for S3 File source

aditya300899:
@shawkins I think this should work, let me know your thoughts! Thanks!


aditya300899:
I added the escape character to "*" as it was giving some dangling metacharacter exception as it is used in a regex.

aditya300899:
This is not needed anymore! Incorporated in #1323

PR #1329, opened 2020-09-03
Parquet refinements

shawkins:
Adds an end to end working example and handling for all predicate pushdown.


connectors/file/translator-parquet/src/main/java/org/teiid/translator/parquet/ParquetQueryVisitor.java
```diff
@@ -37,21 +40,36 @@
...
 
     private List<String> projectedColumnNames = new ArrayList<String>();
-    private LinkedHashSet<String> partitionedColumns = new LinkedHashSet<>();
+    private LinkedHashSet<String> allColumns = new LinkedHashSet<>();
+    private LinkedHashMap<String, Column> partitionedColumns = new LinkedHashMap<>();
+
+    //we separate predicates into three buckets
+    //comparisons that can be applied to the file search path
     private Map<String, Comparison> partitionedComparisons = new HashMap<String, Comparison>();
+    //predicates that can be evaluated solely against the file path once the file is retrieved
```
aditya300899, line 55: I did not particularly understand this. Any example would do.

aditya300899 (approved):
This implementation looks very cool. Though was very overwhelming for me to understand. LGTM.

connectors/file/translator-parquet/src/main/java/org/teiid/translator/parquet/ParquetQueryVisitor.java
shawkins: The predicate logic breaks the where clause down by separating each conjunct.  We can treat conjuncts as logically independent of one another.  For example:

something AND (something1 OR something2) - has 2 conjuncts

For each conjunct we can see if it applies only against the partitioning columns, only against non-partitioning columns, or a mixture of both.  

If the conjunct applies only to the partitioning columns there are two cases:
- it's just an equality comparison, so we can use the logic you developed to create a specific search pattern
- otherwise it's a predicate that we can check against the file path (something like WHERE "year" > 2010) - before you even create the intermediate file copy.  More on this below.

If the conjunct applies only to the non-partitioning columns the logic is effectively the same as what you had - it's turned into a FilterPredicate using the standard FilterApi.

If the conjunct applies to both partitioning and non-partitioning columns (something like WHERE "year" = 2010 OR name = 'Aditya') there is extra work to do.  To evaluate the whole expression we need to convert the partitioning predicates into FilterApi UserDefinedPredicates.  Those UserDefinedPredicates logically evaluate the comparison of partition values to the predicate literal.  However it won't get evaluated properly by the parquet logic unless we associate it with an actual column, so we have to find a reference column - it doesn't really matter which one, just as long as it's of a type that is generally comparable.

Since we have already developed this handling for UserDefinedPredicates for partitioning columns, we can then reuse that to evaluate non-equality comparisons against just the path (from above something like WHERE "year" > 2010) by using an instance of a Visitor that performs the evaluation.  You can see this is checked before the file is handed off for processing.

With all of that in place we are filtering files as much as possible, and there's no need for the usage of partial filtering.

PR #1333, opened 2020-09-28
Fixed wrong table names for geometry metadata tables

treiboll:
Based on exceptions in the log file I found an error in the PostgreSQLMetadataProcessor. The table names for geometry metadata tables are mixed-up.


shawkins:
Thank you, that was certainly a mistake.

PR #1335, opened 2020-10-11
Prepare to multithreaded use of PartnerConnection

3DRaven:
Hi. We had discussion about MT of PartnerConnection. You really use shared state of headers and options. So, for preparing of multithreaded manner of use PartnerConnection I moved this options to constructor. Because this options really always set as constants. And as additional, create of bulk job made sunchronized. 

This made for further add new functionality in spring-boot-teiid project with connection pool.

### DIFF
--- a/connectors/salesforce/translator-salesforce/src/main/java/org/teiid/salesforce/BaseSalesforceConnection.java
+++ b/connectors/salesforce/translator-salesforce/src/main/java/org/teiid/salesforce/BaseSalesforceConnection.java
@@ -95,7 +95,8 @@ public abstract class BaseSalesforceConnection<T extends SalesforceConfiguration
     public BaseSalesforceConnection(T salesforceConfig) throws AsyncApiException, ConnectionException {
         config = createConnectorConfig(salesforceConfig);
         partnerConnection = login(salesforceConfig, config);
-
+        partnerConnection.setMruHeader(false); //Because really MRU used only in query and always false
+        partnerConnection.setQueryOptions(2000);//Because we have hard SF limit and really this value always is 2000
         String endpoint = config.getServiceEndpoint();
         // The endpoint for the Bulk API service is the same as for the normal
         // SOAP uri until the /Soap/ part. From here it's '/async/versionNumber'
@@ -199,44 +200,29 @@ public abstract class BaseSalesforceConnection<T extends SalesforceConfiguration
         return partnerConnection;
     }
 
-    public QueryResult query(String queryString, int batchSize, boolean queryAll) throws TranslatorException {
-
-        if(batchSize > 2000) {
-            batchSize = 2000;
-            LogManager.logDetail(LogConstants.CTX_CONNECTOR, "reduced.batch.size"); //$NON-NLS-1$
-        }
+    @Override
+    public QueryResult query(String queryString, boolean queryAll) throws TranslatorException {
 
         QueryResult qr = null;
-        partnerConnection.setQueryOptions(batchSize);
         try {
             if(queryAll) {
                 qr = partnerConnection.queryAll(queryString);
             } else {
-                partnerConnection.setMruHeader(false);
                 qr = partnerConnection.query(queryString);
             }
         } catch (ConnectionException e) {
             throw new TranslatorException(e);
-        } finally {
-            partnerConnection.clearMruHeader();
-            partnerConnection.clearQueryOptions();
         }
         return qr;
     }
 
-    public QueryResult queryMore(String queryLocator, int batchSize) throws TranslatorException {
-        if(batchSize > 2000) {
-            batchSize = 2000;
-            LogManager.logDetail(LogConstants.CTX_CONNECTOR, "reduced.batch.size"); //$NON-NLS-1$
-        }
-
-        partnerConnection.setQueryOptions(batchSize);
-        try {
+    @Override
+    public QueryResult queryMore(String queryLocator) throws TranslatorException {
+        
+    	try {
             return partnerConnection.queryMore(queryLocator);
         } catch (ConnectionException e) {
             throw new TranslatorException(e);
-        } finally {
-            partnerConnection.clearQueryOptions();
         }
 
     }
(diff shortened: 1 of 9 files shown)


shawkins:
This doesn't quite capture the intention of setting the batch size as it will now always be fixed at 2000.  The engine depending upon the data width of the expected result and the setting for the processing batch size is passing a calculated value.  For particularly wide results, 2000 could be too large of a value.

I'm fine setting the mru to false by default.

The synchronization around the bulkconnection can be avoided by either creating another member bulkconnection specifically for chunking, or to change the logic in the method to create a new bulkconnection for chunking.

This also doesn't address the security related issue - that a single credential will be associated with all usage.

All of this is because you are trying to avoid a connection pool in teiid-spring-boot correct?  Have you looked into adding pooling support, or do you have a hard expectation of using a single connection in a multi-threaded manner?

3DRaven:
> This doesn't quite capture the intention of setting the batch size as it will now always be fixed at 2000. The engine depending upon the data width of the expected result and the setting for the processing batch size is passing a calculated value. For particularly wide results, 2000 could be too large of a value.

Ok, I found calculation of FetchSize in registerRequest. Tnx. 
class RegisterRequestParameter have field fetchSize, this field really used.

https://developer.salesforce.com/docs/atlas.en-us.api.meta/api/sforce_api_header_queryoptions.htm

In SOAP Docs wrote: "changes are sometimes made to maximize performance". Have you are checking about performance when we set batchSize? May be default value as 500 is good choice for this (as good start may be)? Now we are have lot of problems because we used one connection per request in teiid-spring-boot (we made our solution as workaround for this, for maximize performance, but it is just "Ugly hack") and in wildfly connection pool really is not working.

https://issues.redhat.com/browse/TEIID-6041
I checked really resource adapter do not create any connections for pool when wildfly started (init connections and min connections is not work and connections is closing immediately after request).

> The synchronization around the bulkconnection can be avoided by either creating another member bulkconnection specifically for chunking, or to change the logic in the method to create a new bulkconnection for chunking.

Ok, I will create new object BulkConnection if it is normal.

> This also doesn't address the security related issue - that a single credential will be associated with all usage.

Can you describe real scenarios when Teiid try to use multiple credentials for connect to salesforce in one resource adapter (for wildfly) or get multiple credentials when start new connection per request in spring-boot-teiid?

> All of this is because you are trying to avoid a connection pool in teiid-spring-boot correct? Have you looked into adding pooling support, or do you have a hard expectation of using a single connection in a multi-threaded manner?

I think we can use one connection with multithreaded manner and add connections pool too to spring-boot-teiid. In other scenarios we need to make size of pool=number of requests executed in server parallel, this is expensive as I think.

shawkins:
> I checked really resource adapter do not create any connections for pool when wildfly started (init connections and min connections is not work and connections is closing immediately after request).

I think there is some confusion on that issue.  Are you talking about the Teiid SalesforceConnection, or the http connection for the initial login?  It's expected that the latter will close.  If you do a cli operation to get the connection count in the pool you should see the expected connections.

> Can you describe real scenarios when Teiid try to use multiple credentials for connect to salesforce in one resource adapter (for wildfly) or get multiple credentials when start new connection per request in spring-boot-teiid?

Teiid supports credential delegation from the user login to the underlying connection pool.  With the use of login modules you can do quite a bit in that regard.  That is not currently supported by Teiid Spring Boot - there's no delegation nor login modules in play.  At a minimum the goal is to keep any logic that presumes a single login for all connections out of Teiid core.

3DRaven:
>  If you do a cli operation to get the connection count in the pool you should see the expected connections.

I spend lot of time for search this CLI commands, but I got only "resource not found" for all commands for pool statistics
{code}
/subsystem=resource-adapters/resource-adapter=salesforcefour:write-attribute(name=statistics-enabled,value=true)
/subsystem=resource-adapters/resource-adapter=salesforcefour/connection-definitions=sfDS41:read-resource(include-runtime=true)
/subsystem=resource-adapters/statistics=statistics/connection-definitions=java\:\/sfDS41:read-resource(include-runtime=true)
{code}

this commands was not work.

Can you take commands please for checking?

> Teiid supports credential delegation from the user login to the underlying connection pool. With the use of login modules you can do quite a bit in that regard. That is not currently supported by Teiid Spring Boot - there's no delegation nor login modules in play. At a minimum the goal is to keep any logic that presumes a single login for all connections out of Teiid core.

Ok, as I understand I can make multi-thread connection using with same credentials at SalesforceConnection level? 
So, plan:
**Phase 1**
1. Change synchronized(this) to new object creation.
2. Set batchSize to default value (500 in documentation) in connection, but do not change fetchSize calculation in engine (for futher development).
3. Use SalesforceConnection in multi-thread manner with one credentials.
**Phase 2**
1. Make melti-threaded reusable single connection in spring-boot-teiid.
**Phase 3**
1. Make multithreaded reusable connections pool in spring-boo-teiid.

As you think?

shawkins:
> Can you take commands please for checking?

/subsystem=resource-adapters/resource-adapter=salesforcefour:write-attribute(name=statistics-enabled,value=true)

Is roughly how that would be set for JDBC.  For other resource adapters it is specified on the pool for the given connection definition:

/subsystem=resource-adapters/resource-adapter=salesforcefour/connection-definitions=sfDS/statistics=pool:write-attribute(name=statistics-enabled, value=true)

You just need to substitute in the name of the connection definition you used - that is the logical name, which is likely not the jndi name.  You can auto complete on subsystem=resource-adapters/resource-adapter=salesforcefour/connection-definitions= to see your connection definition names.  You should then be able to read the pool statistics against that connection definition name.

> Set batchSize to default value (500 in documentation) in connection, but do not change fetchSize calculation in engine (for futher development).

That's not a great solution.  Also you would use a connection in a multi-threaded manner, or via a connection pool - you don't need to do both.  

The core issue here is the statefulness of the salesforce api and the usage pattern with Teiid.  It seems simpler to approach this in a similar way as the bulk connection logic - just create new partner connection when needed to account for batch sizes and other stateful information.  

If you take the connector config post login that has the session id and service url set and mark it as manual login:

config.setManualLogin(true);

Then it can be passed to a new PartnerConnection and avoid the remote call to login.  Does that make sense?

shawkins:
Closing and replacing with https://issues.redhat.com/browse/TEIID-6052 which will be implemented as described in the last comment.

3DRaven:
I made this in myself code in app by replace Teiid beans. But this is just hack because I had some problems with discussion about this PR :) But now I do not need this workaround, tnx :)

PR #1339, opened 2021-01-29
Order by fix

3DRaven:
This is fix of https://issues.redhat.com/browse/TEIID-6069


shawkins:
Thanks that looks good.

PR #1348, opened 2021-04-14
First push of druid translator

dkrapohl:
We're using this in production right now so it works in Teiid 10.3.X with Avatica 0.17 jdbc driver.

Capturing for public consumption:  

1. Put Avatica driver at the path modules\system\layers\base\org\apache\calcite\main\
2. Add module.xml with content
 `<module xmlns="urn:jboss:module:1.0" name="org.apache.calcite">
  <resources>
    <resource-root path="avatica-1.17.0.jar"/>
  </resources>
  <dependencies>
    <module name="javax.api"/>
  </dependencies>
</module>`
3. Add to standalone-(whichever you use).xml
`				<datasource .......>
                    <connection-url>jdbc:avatica:remote:url=(YOUR CONNECTION STRING)</connection-url>
                    <driver>org.apache.calcite</driver>
                </datasource>	`
4. Add to standalone-(whichever you use).xml
`<translator name="druid" module="org.jboss.teiid.translator.jdbc"/>`

### DIFF
--- a/connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/druid/DruidExecutionFactory.java
+++ b/connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/druid/DruidExecutionFactory.java
@@ -1,0 +1,448 @@
+/*
+ * Copyright Red Hat, Inc. and/or its affiliates
+ * and other contributors as indicated by the @author tags and
+ * the COPYRIGHT.txt file distributed with this work.
+ *
+ * Licensed under the Apache License, Version 2.0 (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ * http://www.apache.org/licenses/LICENSE-2.0
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.teiid.translator.jdbc.druid;
+
+import org.teiid.language.*;
+import org.teiid.metadata.RuntimeMetadata;
+import org.teiid.translator.*;
+import org.teiid.translator.jdbc.*;
+import org.teiid.translator.jdbc.oracle.ConcatFunctionModifier;
+import org.teiid.translator.jdbc.oracle.OracleFormatFunctionModifier;
+
+import java.sql.Connection;
+import java.sql.Time;
+import java.sql.Timestamp;
+import java.util.*;
+
+import static org.teiid.translator.TypeFacility.RUNTIME_NAMES.*;
+
+/**
+ * Translator class for Apache Druid.
+ * Created by Don Krapohl 04/02/2021
+ */
+@Translator(name="druid", description="druid custom translator")
+public class DruidExecutionFactory extends JDBCExecutionFactory {
+
+    public static String DRUID = "druid"; //$NON-NLS-1$
+
+    static final String UUID_TYPE = "uuid"; //$NON-NLS-1$
+    private static final String INTEGER_TYPE = "integer"; //$NON-NLS-1$
+    private String druidBrokerURL;   //required, defaults to null
+    private String serialization="json"; //defaults to json
+    private String authentication="NONE";    //defaults to no authentication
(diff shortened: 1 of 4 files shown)


shawkins:
Thanks @dkrapohl  The build is complaining about very minor things - like line endings.  I don't mind at all if this gets included in the main project source.  We can convert what you have in the description to a doc page.

shawkins:
Would be able to resolve the rest of the checkstyle and other minor build errors?  We can set a goal of having a Teiid 17 in July with this included.

dkrapohl:
Can do.  I'll get this out next week.

connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/DefaultSQLDialect.java
```diff
@@ -1,34 +1 @@
-/*
```
shawkins, line 1: The line endings here seem to have been mangled.

connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/druid/DruidExecutionFactory.java
```diff
@@ -0,0 +1,531 @@
...
+
+    /**
+     * Postgres doesn't provide min/max(boolean), so this conversion writes a min(BooleanValue) as
+     * bool_and(BooleanValue)
+     * @see org.teiid.language.visitor.LanguageObjectVisitor#visit(org.teiid.language.AggregateFunction)
+     * @since 4.3
+     */
+    @Override
+    public List<?> translate(LanguageObject obj, ExecutionContext context) {
+        if (obj.getClass().toString().equals("class org.teiid.language.Select")) {
```
shawkins, line 305: Why a string name comparison and not an instanceof check?

connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/druid/DruidExecutionFactory.java
```diff
@@ -0,0 +1,531 @@
...
+            if (null!= ((Select) obj).getGroupBy() && ((Select) obj).getGroupBy().getElements().size() > 0) {
+                // remove any columns from group by that are not in the select
+                ((Select) obj).getGroupBy().getElements().removeAll(
```
shawkins, line 308: You probably don't want to do this here - the projected columns won't line up and you should get exceptions later on.  You probably need a capability similar to supportsOrderByUnrelated - where the engine won't allow for grouping expressions that don't appear in the select.

connectors/jdbc/translator-jdbc/src/test/java/org/teiid/translator/jdbc/druid/TestDruidTranslator.java
```diff
@@ -0,0 +1,149 @@
...
+        input = "select TIMESTAMPADD(SQL_TSI_DAY, 1, timestampvalue) from BQT1.MediumA"; //$NON-NLS-1$
+        output = "SELECT MediumA.TimestampValue + (INTERVAL '1' DAY(1)) FROM MediumA"; //$NON-NLS-1$
+        helpTestVisitor(getTestBQTVDB(), input, output);
+
+        //input = "select TIMESTAMPDIFF(SQL_TSI_DAY, NOW(), timestampvalue) from BQT1.MediumA"; //$NON-NLS-1$
+        //output = "SELECT MediumA.TimestampValue + (INTERVAL '1' DAY(1)) FROM MediumA"; //$NON-NLS-1$
+        //helpTestVisitor(getTestBQTVDB(), input, output);
+    }
+    @Test
+    public void testStringToTimestamp() throws Exception {
```
shawkins, line 124: This test fails for me locally:

Actual result: TIME_PARSE('2004-06-29 23:59:59.987', 'yyyy-MM-dd HH:mm:ss')

connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/druid/DruidExecutionFactory.java
dkrapohl: Ouch.  Fixed and pushed.

connectors/jdbc/translator-jdbc/src/main/java/org/teiid/translator/jdbc/DefaultSQLDialect.java
dkrapohl: IDE on windows.  Fixed and pushed.

